PR #41, opened 2024-10-07
fix (gradle): refactor number of files to change instances of sample to bookswap

oliveja22:
Includes a number of refactored instances of sample to bookswap in order to have everything under a proper name and not its standard downloaded state.


LeniCadorinEpfl (approved):
Every instance of "sample" were correctly changed to bookswap.
Everything is correct for me.

PR #44, opened 2024-10-08
Add representation of books in firebase and in the app

matiisalv:
-Data Class(Databook): represent a book with different attributes,Title(String),Author(String),Description(String),Rating(Int),Language(enum),ISBN(String) -Enum Language:Represent the languages of the app
-Interface(Bookrepository) : representing the management of Book's data -Class (BooksFirestoryRepository):Implement the inferface(Bookrepository) and manage the communication between the app and firestore for the books
-Include all tests


app/src/main/java/com/android/bookswap/model/BooksFirestoreRepository.kt
```diff
@@ -0,0 +1,101 @@
...
+  // Maps a Firestore document to a DataBook object
+  // If any required field is missing, returns null to avoid incomplete objects
+  fun documenttoBooks(document: DocumentSnapshot): DataBook? {
+    return try {
+      val title = document.getString("Title") ?: return null
+      val author = document.getString("Author") ?: return null
+      val description = document.getString("Description") ?: return null
+      val rating = document.getString("Rating") ?: return null
+      val photo = document.getString("Photo") ?: return null
+      val language = document.getString("Language") ?: return null
```
oliveja22, line 78: Here the use of two language variables, one for the string and the other after being transformed to the type feels unnecessary, maybe mix the two to have a single variable, even maybe a small helper function that transforms it if necessary.

oliveja22 (changes requested):
Be careful, I am sure they work, but there are two different instances of a file that tests, they seem somewhat similar but one seems more developed than the other. I am guessing this is not intentional. 
You are also missing the tests for inti() and getNewUid(). They are more simple but nonetheless deserve tests.

matiisalv:
The init test is not there as it requires the implementation of the authentication which is not yet done.

PR #46, opened 2024-10-08
Add geolocation to the app

Chockencore:
This PR adds the possibility of finding the current user's location with the geolocation of the android phone with the help of the google play services location library.
The geolocation must be firstly created by giving it the current Activity of the screen, then to find the user's location, we must use the startLocationUpdates function. If given the permission by the phone, it will update  automatically the user's position which is accessible with the longitude and latitude mutableState. Otherwise, it will ask the user the permission to access the user's geolocation.
To  stop the tracking of the phone's geolocation, we use the stopLocationUpdates function. When using this class, it should be remembered to always stop the geolocation when exiting a screen that uses the geolocation unless we have a reason to still track the geolocation of the user in the background.


Chockencore:
The SonarCloud analysis didn't work because of the permission the program ask in the manifest. These permissions need to be checked to be safe for the user on SonarCloud.  If the person that will review my PR can also go check if these permisions are truly necessary and mark them as safe or tell me which one I should remove, it would be appreciated.

app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
```diff
@@ -0,0 +1,74 @@
...
+      LocationRequest.create().apply {
+        interval = 10000 // Update interval in milliseconds
+        fastestInterval = 5000 // Fastest update interval in milliseconds
+        priority = LocationRequest.PRIORITY_HIGH_ACCURACY
+      }
+
+  private val locationCallback =
+      object : LocationCallback() {
+        override fun onLocationResult(p0: LocationResult) {
+          p0.lastLocation.let { location ->
```
TheoSchlaeppi, line 30: Potential Issue: According to the Android documentation, lastLocation may return null in certain situations (e.g., when location services are disabled, the device hasn't received a location fix, or the user denies permissions). This can lead to a NullPointerException

app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
```diff
@@ -0,0 +1,74 @@
...
+    fusedLocationClient = LocationServices.getFusedLocationProviderClient(activity)
+  }
+
+  // Request location permissions
+  private fun requestLocationPermissions() {
+    val permissions =
+        arrayOf(
+            Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_COARSE_LOCATION,
+            Manifest.permission.ACCESS_BACKGROUND_LOCATION)
```
TheoSchlaeppi, line 49: Potential issue : ACCESS_BACKGROUND_LOCATION needs API LEVEL 29 which is not the case for all devices.  I find a stackOverflow that can help you for this issue [here](https://stackoverflow.com/questions/73034972/why-does-the-permissions-dialog-not-appear-when-i-ask-for-background-location)

TheoSchlaeppi (changes requested):
In general, your code is well-structured and well-written.
However, there are a couple of issues that I'd like to highlight for improvement:

1. Risk of NullPointerException (lastLocation):
**Issue**: In the current code, you're accessing latitude and longitude from lastLocation without checking if lastLocation is null.
**Suggestion**: I recommend adding a null check for lastLocation before accessing its properties.

2. Handling Background Location Permission (API 29+):
**Issue**: On Android 10 (API level 29) and higher, ACCESS_BACKGROUND_LOCATION must be requested separately from foreground location permissions (ACCESS_FINE_LOCATION and ACCESS_COARSE_LOCATION). In the current code, all three permissions are requested at once, which could lead to issues on Android 10+ devices.
**Suggestion**: You should only request ACCESS_BACKGROUND_LOCATION after the user has granted foreground location permissions. Additionally, you can use the @RequiresApi annotation to ensure that background location permission requests are only made on devices running Android 10 or higher.

3. SonarCloud Analysis:
I checked the SonarCloud report, and it seems that all three location access permissions (ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION, and ACCESS_BACKGROUND_LOCATION) are necessary for the functionality you're aiming for.

Great work overall! Let me know if you have any questions or need further clarification.

LeniCadorinEpfl (commented):
I agree with what Theo mentioned, your code is well-structured and well-written.
I think the first two points Theo highlighted are important and can be changed for an even better code.
 Here are a few additional points I'd like to propose for improving the code:
 1.  MutableState vs. MutableDoubleState:
 Issue: In the code, you're using mutableDoubleStateOf for latitude and longitude, this is fine. But using the MutableState type may offer more flexibility and I think is better suited for Compose-based updates.
Suggestion: I recommend refactoring this to mutableStateOf<Double>()
2. Testing for Null LastLocation and Permission Denial:
Issue: Looking at the tests, which I think are very well done indeed. I think additional tests could be added to improve coverage:
2.1 : Null Handling for lastLocation: Add a test to check the correct behaviour in the case when lastLocation returns null. This follows a bit the point suggested by Theo. 
Suggestion : Mock lastLocation to return null and verify the behvaiour.
2.2 : Permissions Denial: Add a test to check the behavior when permissions are denied and how the app handles retrying for location updates.
Suggestion: Verify how the app behaves when location permissions are denied.
Adding these test cases will further ensure the robustness of your implementation, especially for edge cases.

Overall, I'd say the code is great. These are just points to make it even better. Also if any of the comments are unclear just let me know.

app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
```diff
@@ -0,0 +1,114 @@
...
+  @RequiresApi(Build.VERSION_CODES.Q)
+  private fun hasBackgroundPermissions(): Boolean {
+    return ActivityCompat.checkSelfPermission(
+        activity, Manifest.permission.ACCESS_BACKGROUND_LOCATION) ==
+        PackageManager.PERMISSION_GRANTED
+  }
+
+  // Start location updates
+  @SuppressLint("MissingPermission")
+  fun startLocationUpdates() {
```
TheoSchlaeppi, line 87: I don't know if it will be possible in the future, but there's a chance that startLocationUpdates might be called even when it’s already running. You should either check if it's running within this function or ensure the check is done before calling it

TheoSchlaeppi (commented):
The issues from the initial request have been addressed. The new sections are nearly perfect. However, in startLocationUpdates, there is duplicated code. In my opinion, a boolean flag should be used to track if the updates are already running, to avoid redundant calls.
Great work overall, you're almost finished!

TheoSchlaeppi (approved):
Looks good for me. Well done !

PR #48, opened 2024-10-09
Implement UI that works with book addition

matiisalv:
⁠Create a view class ⁠ listToBooksView ⁠ to manage book interactions with the repository.
•⁠  ⁠Add methods to fetch books, add new books, and generate unique identifiers.
•⁠  ⁠Implement ⁠ AddToBook ⁠ Composable to provide input fields for book details (title, author, description, rating, ISBN, photo, language).
•⁠  ⁠Validate user input before adding a book and provide feedback via Toast messages.
•⁠  ⁠Ensure the UI updates reactively with the StateFlow for book lists and selected books.


app/src/main/java/com/android/bookswap/ui/Addbook/AddToBook.kt
```diff
@@ -0,0 +1,206 @@
...
+  // Scaffold to provide basic UI structure with a top app bar
+  Scaffold(
+      modifier = Modifier.testTag("addBookScreen"),
+      topBar = {
+        TopAppBar(
+            // Title of the screen
+            title = { Text("Add Your Book", modifier = Modifier.testTag("addBookTitle")) },
+            // Icon button for navigation (currently no action defined)
+            navigationIcon = {
+              IconButton(onClick = {}) {
```
oliveja22, line 49: I know it is a little useless to mention, but I think it would be better to leave the button out if it has no implementation yet. It might let the person who does it have more freedom.

oliveja22 (commented):
There are a number of things that I would like to mention, though none of them are crucial as far as I can tell.
The first is how you check whether the fields are filled in correctly or not. At the moment, you just check if they are left empty or not. This is not enough I believe, after all, many of those have quite a few restrictions. The main one is the ISBN, which has a number of restrictions on its structure which should be reflected. This then should be tested I believe.
Additionally, there seems to be very few tags when it comes to the modifiers oof the Outlined Text Fields. They should not resemble what is present in the figma, they do not have rounded corners, nor do they match the colour scheme. 
As mentioned though, these are light problems that I do not see having major consequences for the moment.
 Even more useless I added a comment in the code, it is minute and frankly debatable, so feel free to ignore it if you believe it to be wrong, it can mostly be summed up to preference after all.

app/src/main/java/com/android/bookswap/ui/Addbook/AddToBook.kt
```diff
@@ -0,0 +1,291 @@
...
+      })
+}
+
+fun createDataBook(
+    title: String,
+    author: String,
+    description: String,
+    ratingStr: String,
+    photo: String,
+    Book_languageStr: String,
```
EdenKahane, line 233: Variables should always follow lower camel case

app/src/main/java/com/android/bookswap/ui/Addbook/listbookview.kt
```diff
@@ -0,0 +1,52 @@
+package com.android.bookswap.ui.Addbook
+
+import com.android.bookswap.data.DataBook
+import com.android.bookswap.model.Booksrepository
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+// A view class that interacts with the Books repository to manage a list of books.
+open class listToBooksView(private val repository: Booksrepository) {
```
EdenKahane, line 10: Class name should be upper camel case

app/build.gradle.kts
```diff
@@ -131,11 +136,15 @@ dependencies {
...
     implementation(libs.androidx.navigation.compose)
+    implementation(libs.firebase.crashlytics.buildtools)
+    implementation(libs.identity.jvm)
+
     testImplementation(libs.junit)
     testImplementation(libs.mockito.core)
     testImplementation(libs.mockito.kotlin)
     testImplementation(libs.mockk)
     testImplementation(libs.mockito.inline)
+    implementation("org.bouncycastle:bcprov-jdk18on:1.78.1")
```
EdenKahane, line 147: "org.bouncycastle:bcprov-jdk18on:1.78.1" need to be in libs.versions.toml

app/src/androidTest/java/com/android/bookswapp/ui/addbooktest.kt
```diff
@@ -0,0 +1,235 @@
...
+}
+
+class MockBooksRepository : Booksrepository {
+  var isBookAdded = false
+  var isBookFetched = false
+  var isBookUpdated = false
+  var isBookDeleted = false
+  var shouldFail = false
+
+  override fun init(OnSucess: () -> Unit) {
```
EdenKahane, line 181: rename to onSucess, otherwise it's perfect !

app/src/main/java/com/android/bookswap/ui/Addbook/listbookview.kt
EdenKahane, line 1: This is a complex interaction with the data layer. So it's in the model layer not the UI.
Just move this to the model folder and :ok_hand:

app/src/main/java/com/android/bookswap/ui/Addbook/listbookview.kt
```diff
@@ -0,0 +1,52 @@
...
+  // selected book.
+  open val selectbook: StateFlow<DataBook?> = selectbook_.asStateFlow()
+
+  // Initialization block. Calls the repository's init() function when the class is instantiated.
+  init {
+    repository.init {}
+  }
+
+  // Function to generate a new unique identifier (UID) for a book, using the repository.
+  fun getNewUid(): String {
```
EdenKahane, line 31: We're currently using both Uid as string and UUID as identifier, this is a problem.
Everything should be UUID (firestore support this without a problem).
We will need to refactor this when we refactor firestore

EdenKahane (changes requested):
Very good code, there is some minor name convention problem and some files need to be moved to follow app architecture.
Apart from this, it's very clean ! 

We will need at some point to merge the Uid system and the UUID system, this is not the goal of  this task but it will not be fun.

app/src/main/java/com/android/bookswap/ui/Addbook/AddToBook.kt
```diff
@@ -0,0 +1,291 @@
...
+        // Column layout to stack input fields vertically with spacing
+        Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+          // Title Input Field
+          OutlinedTextField(
+              value = title,
+              onValueChange = { title = it },
+              label = { Text("Title") },
+              placeholder = { Text("Enter the book title") },
+              modifier =
+                  Modifier.padding(paddingValues)
```
EdenKahane, line 80: paddingValues is only for root component of the content (so Column  here)

matiisalv:
Thank you for review, as we talk about , I am currently adding this , I will deliver it soon

matiisalv:
The code have problem with a particular test that doesn't pass in the CI but yes in my own device

app/src/test/java/com/android/sample/ExampleUnitTest.kt
EdenKahane, line 1: This file should not be added (you need to remove it from your app local branch).

app/src/test/java/com/android/sample/PointTest.kt
EdenKahane, line 1: Also do not add this file

gradle/libs.versions.toml
```diff
@@ -21,23 +21,26 @@ robolectric = "4.11.1"
 sonar = "4.4.1.3373"
 mockk = "1.12.0"
 firebaseFirestoreKtx = "25.1.0"
-
+json = "20210307"
```
EdenKahane, line 24: Duplication, json still exist at the line 45 of the file

app/src/main/java/com/android/bookswap/MainActivity.kt
EdenKahane, line 1: You should not modify this file back, keep the new code of the main branch (rollback the change on your local branch)

app/src/androidTest/java/com/android/bookswapp/ui/addbooktest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+// Simulating the listToBooksView interface implementation
+class MockListToBooksView(repository: Booksrepository) : listToBooksView(repository) {
+  var isBookAdded = false
+  var addedBook: DataBook? = null
+  var isBookDeleted = false
+  var isBookUpdated = false
+  var fetchedBooks: List<DataBook> = emptyList()
+
+  override fun Add_Book(book: DataBook) {
```
EdenKahane, line 132: Not following naming convention for methods

app/src/androidTest/java/com/android/bookswapp/ui/addbooktest.kt
EdenKahane: Now with the new app refactoring, see new [function method](https://github.com/BookswapEPFL/Bookswap/blob/main/app/src/main/java/com/android/bookswap/data/repository/BookRepository.kt)
These one are deprecated

EdenKahane (changes requested):
Secrets leak ! Need to be remove (not just from files, the history of the commit need to be reset !).
Previous requested changes are still here and not fixed
The app should follow the new convention, especially for the new methods of BookRepository.

PR #49, opened 2024-10-10
fix!: Fixed DataBook name convention and data types

EdenKahane:
Breaking change !  This fix is quite important since the amount of time we're supposed to use this class.

Fixed DataBook data class :
- Moved DataBook.kt to data folder instead of model folder since it's not a model class 
- Variables name should always start with a lowercase letter
- Function/Class comments should be between /**   */ to be shown by Android Studio description helper
- Class description of each parameter should be in the comment
- Rating should be optional as isbn provider doesn't give a rating
- language renamed to BookLanguages so people can't confuse it for the app language
- OTHER enum value added for BookLanguages for any book not in the coded language.


Chockencore (changes requested):
This pull request is great but the ISBN of a book may not exist if it is a self published book as such it should be optional.

Chockencore (changes requested):
The books shouldn't be searched by its title like in the addbooks or updatebook nor by its ISBN like in deletebooks as certain book may not have an ISBN. Also the name of the function should be consistent, it should be addBook and not addBooks, and the first letter of the second noun in the function should be a capital letter, so updatebook into updateBook.

EdenKahane:
> The books shouldn't be searched by its title like in the addbooks or updatebook nor by its ISBN like in deletebooks as certain book may not have an ISBN. Also the name of the function should be consistent, it should be addBook and not addBooks, and the first letter of the second noun in the function should be a capital letter, so updatebook into updateBook.

This files were not created by me and need to be heavily edited (see #63 ), as the code need to be redone anyway for these two files, I just made the change so we could continue with the new data type.

Chockencore (approved):
I approve this PR as it is urgent to have the modification in main, the requested change will be addressed in another issue/task

EdenKahane:
The other fixes needed are in #63

PR #50, opened 2024-10-10
feature: Add request of books from GoogleBook using ISBN

### LINKED ISSUE #39: [2.1] Make a method to request info from API using ISBN for book

EdenKahane:
Add methods to request and parse a book from Google cloud with security checks.
Tests for the parsing part are implemented.

Close #39


EdenKahane:
Waiting for #49

app/src/main/java/com/android/bookswap/data/source/api/GoogleBookDataSource.kt
```diff
@@ -0,0 +1,73 @@
...
+      // we do not always use get..OrNull()
+      val item = json.getJSONArray("items").getJSONObject(0).getJSONObject("volumeInfo")
+
+      val language =
+          when (val languageCode = item.getStringOrNull("language")?.uppercase()) {
+            is String -> Languages.values().first { it.languageCode == languageCode }
+            else -> null
+          }
+      return Result.success(
+          DataBook(
```
EdenKahane, line 61: Waiting for #49 to use the new system

EdenKahane (commented):
Waiting for #49

EdenKahane:
> The code is great, there is only minor details to change.
> 
>     1. The parseISBNResponse function should be private as it have the tag @VisibleForTesting and it should only be used by getBookFromISBN.
> 
>     2. The ISBN should be only 13 digits long. In the past the ISBN was 10 digits but with the current system, only the 13 digits exist. The convention says that we need to put 978 before the old ISBN to have the current ISBN then we recompute the check digit. This transformation for old books should occur at the creation of the book, so the code should assume that the book has the current version of the ISBN and check that the ISBN is 13 digits long.



> The code is great, there is only minor details to change.
> 
>     1. The parseISBNResponse function should be private as it have the tag @VisibleForTesting and it should only be used by getBookFromISBN.
> 
>     2. The ISBN should be only 13 digits long. In the past the ISBN was 10 digits but with the current system, only the 13 digits exist. The convention says that we need to put 978 before the old ISBN to have the current ISBN then we recompute the check digit. This transformation for old books should occur at the creation of the book, so the code should assume that the book has the current version of the ISBN and check that the ISBN is 13 digits long.

1. Set otherwise tag, but I can't change the function to private or it won't work
2. Now fixed

Chockencore (changes requested):
The code now add the 978 at the beginning of the ISBN of 10 digits but like I said in the last review, the code needs to recompute the check digits, the leftmost digit as it may not be the same as previously.
Here is the computation for the last digit :

    Take the first 12 digits of the 13-digit ISBN
    Multiply each number in turn, from left to right by a number. The first digit is multiplied by 1, the second by 3, the third by 1 gain, the fourth by 3 again, and so on to the eleventh which is multiplied by 1 and the twelfth by 3.
    Add all of the 12 answers.
    Do a modulo 10 division on the result from step 2. (Don't know what a modulo 10 division is? It's easy. It's just the remainder from a whole number division by 10. I bet you learned to do that in junior school, before you even learned about decimal fractions.)
    Take that remainder result from step 4.If it's a zero, then the check digit is zero. If the remainders isn't zero then subtract the remainder from 10. The answer to that is your check digit.

For example this ISBN : 9781861972712 have 2 as the check digit because :
9*1+7*3+8*1+1*3+...+1*3 = 118  => 118 mod(10) = 8  => 8=/=0 => 10-8 = 2

Chockencore (approved):
I approve this pull request as the request for the google book API accept ISBN 10 and 13 and do the transformation from ISBN 10 to 13 automatically.

PR #51, opened 2024-10-10
Implement the Google Sign-In

LeniCadorinEpfl:
This PR introduces a new screen, the **SignIn screen**. It uses Firebase Authentication and allows users to log in to the app using their Google account.

**Key changes:**

**Sign In Screen UI:**
Implemented the title and logo using composable functions.

**Google Sign-In Button:**
Using Firebase Authentication, it enables connection to a Google account.

**Future Navigation:**
Once navigation is implemented, the commented part can be added and slightly adjusted. After the user signs in, they should be navigated to the Map screen.

**Checklist:**
- [x] CI passes
- [x] Structure is correct


matiisalv (approved):
Brillant work , very clear  and well designed .
In the rememberFirebaseAuthLauncher, the exception handling only logs the error with Log.e. You could provide more user feedback within the UI to notify users about login issues directly:
Toast.makeText(context, "Authentication failed. Please try again.", Toast.LENGTH_LONG).show()
And maybe you should change the name of the testAg to be more specific instead of repeating it 
Two test don't pass please see to it

.github/workflows/ci.yml
```diff
@@ -99,6 +109,10 @@ jobs:
           # To run the CI with debug information, add --info
           ./gradlew check --parallel --build-cache
 
+      # Wait for emulator to fully start
+      - name: Wait for emulator to fully start
```
EdenKahane, line 113: This cause an infinite time loop apparently

EdenKahane (changes requested):
Apart from the weird CI bug, it's perfect !

EdenKahane (approved):
Everything seems fixed, congratulation for fixing the CI problem !

PR #64, opened 2024-10-10
Implements chat feature that takes two users and displays the chat with a possibility of sending messages.

oliveja22:
The UI implemented is a detailed UI with message bubbles in order including a time as well, either an hour and minute if it is within 24 hours or a date otherwise. They alternate sides based on user that sent it as well as colour. It includes a bar at the bottom to write a message that adapts based on size of text and a send button. It updates on the spot as it has a listener that updates the page when there is a change in the database.

It also includes the firebase side of things with send and get functions to obtain necessary functionality as well as the tests for them.


EdenKahane (changes requested):
Some name and emplacement refactoring todo.
And add a UUID system.

app/src/main/java/com/android/bookswap/data/chat/MessageRepository.kt
```diff
@@ -0,0 +1,31 @@
+package com.android.bookswap.data.chat
+
+interface MessageRepository {
+
+  /** Generates a new unique id for a message */
+  fun getNewUid(): String
```
EdenKahane, line 6: use UUID instead of Uid

PR #71, opened 2024-10-10
Add user profile screen, userVM and user data class

Chocolat2:
Add user profile screen and edit profile alert dialogue
Add user data class and view model
Add basic tests for user data class


app/src/main/java/com/android/bookswap/model/userViewModel.kt
```diff
@@ -0,0 +1,175 @@
...
+  open fun getUser(force: Boolean = false): User {
+    if (!isLoaded || force) {
+      fetchUser()
+    }
+    return user
+  }
+
+  private fun fetchUser() {
+    /*
+    firebaseConnection.getUserUidByEmail(user.email).addOnSuccessListener {
```
EdenKahane, line 22: All the firebase part are missing ?

app/src/test/java/com/android/bookswap/data/UserTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    assertEquals(u1.lastName, "")
+    assertEquals(u1.email, "")
+    assertEquals(u1.phoneNumber, "")
+    assertEquals(u1.profilePictureUrl, "")
+    assertEquals(u1.userId, "")
+    assertEquals(u1.address.toStr(), Address(Locale.getDefault()).toString())
+  }
+
+  @Test
+  fun checkAssign() {
```
EdenKahane, line 25: What is the purpose of this test ?

EdenKahane (changes requested):
The code is very clean ! 
The firebase part is commented out for unknown reason 
There is no firebase tests implemented. [Here](https://github.com/BookswapEPFL/Bookswap/blob/main/app/src/test/java/com/android/sample/model/BooksFirestoreRepositoryTest.kt) is an example of firestore mock that should be helpful.

app/src/main/java/com/android/bookswap/model/userViewModel.kt
```diff
@@ -0,0 +1,174 @@
...
+      fetchUser()
+    }
+    return dataUser
+  }
+
+  private fun fetchUser() {
+    /*
+    firebaseConnection.getUserUidByEmail(user.email).addOnSuccessListener {
+        if (it.documents.isNotEmpty()) {
+            this.uid = it.documents[0].id
```
oliveja22, line 24: If you are going to comment so much stuff, I think it would be better to remove it, it makes reading anything a lot more complicated.

oliveja22: I thought it was all finished, please ignore this

app/src/main/java/com/android/bookswap/data/DataUser.kt
```diff
@@ -0,0 +1,30 @@
...
+    var firstName: String = "",
+    var lastName: String = "",
+    var email: String = "",
+    var phoneNumber: String = "",
+    var latitude: Double = 0.0,
+    var longitude: Double = 0.0,
+    var profilePictureUrl: String = "",
+    var userId: String = ""
+) {
+  override fun toString(): String {
```
EdenKahane, line 14: With kotlin dataclass, there is no need to override this

app/src/main/java/com/android/bookswap/data/DataUser.kt
```diff
@@ -0,0 +1,30 @@
...
+    return "$greeting $firstName $lastName"
+  }
+
+  fun printFull1Line(): String {
```
EdenKahane, line 18: This is also unnecessary

EdenKahane (changes requested):
Clean code ! There is a slight unnecessary string function for a data class.
Also next time, please write 2 PR of the two tasks, instead of just one big PR

EdenKahane (approved):
Now everything seems perfect.
Thanks !

PR #73, opened 2024-10-10
Add map screen displaying google maps and books with the help of marker

Chockencore:
This pull request adds a map screen that uses google maps and displays the books that are at the same location with a marker. When clicking on the marker, a box listing all the books title and author at this place appears. Currently the MapScreen takes a list of user, but it is meant to be replaced by a class that manages the interaction with the firebase database to retreive all the user and books. Also this file uses a temporary definition of a user as the definitive version of the DataClass of User is not yet done.
The current user's library that is displayed on the map can be decided at the start of the MapScreen because the marker of googleMaps cannot have a testTag, therefore it is impossible to simulate a click on a marker. Without this argument, the MapScreen would be hard to test.
Screenshot from 2024-10-15 14-23-59


app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    composeTestRule.onNodeWithTag("mapBoxMarker").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("mapBoxMarkerList").assertIsDisplayed()
+    composeTestRule.onAllNodesWithTag("mapBoxMarkerListBox").assertCountEquals(3)
+    composeTestRule.onAllNodesWithTag("mapBoxMarkerListBoxTitle").assertCountEquals(3)
+    composeTestRule.onAllNodesWithTag("mapBoxMarkerListBoxAuthor").assertCountEquals(3)
+    composeTestRule.onAllNodesWithTag("mapBoxMarkerListDivider").assertCountEquals(2)
+  }
+
+  @Test
+  fun noMarkerDisplayedForUserWithoutBooks() {
```
EdenKahane, line 71: Don't hesitate to use full sentence for test function name : 
```kotlin
@Test
fun `no marker is displayed when there is users without books`()  {
}
```

app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -0,0 +1,188 @@
...
+import kotlin.math.roundToInt
+import kotlinx.coroutines.launch
+
+/**
+ * Composable function to display a map with user locations and associated book information.
+ *
+ * This screen renders a GoogleMap that shows books locations as markers. Upon clicking a marker, it
+ * displays a custom info window with the list of books at this location.
+ *
+ * @param listUser List of users (`TempUser`) to display on the map, each containing their location
```
EdenKahane, line 46: In Koltin, for the description, you can use [TempUser] to link a class cleanly (automatically add a link when looking at it in the IDE).

app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -0,0 +1,188 @@
...
+ */
+@Composable
+fun CustomInfoWindow(modifier: Modifier = Modifier, user: TempUser) {
+  Card(
+      modifier =
+          modifier
+              .wrapContentSize()
+              .width(200.dp)
+              .border(
+                  BorderStroke(width = 1.dp, color = Color.hsl(23f, 0.17f, 0.36f, 1f)),
```
EdenKahane, line 151: Avoid using magic colors

EdenKahane (changes requested):
It's great ! 
I have very minor comments but it's mostly for dev visual.

Add a screen of the end result in the PR would also be nice.

app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
Chockencore: I cannot do that as I get an error if I do that in a file in the androidTest folder. I get this error :
"Identifier not allowed in Android projects"
I checked and it also do that in the bootcamp for the androidTest folder. I don't want to spend time to find why it is like this as this is a very minor problem

PR #74, opened 2024-10-11
feature: Create UI screen to create a book from ISBN

EdenKahane:
Add Screen UI for the page that let the user add a book based on a isbn.
This PR use methods of #50 and add configurable UI components that can be used in other pages

- [x] Add tests


EdenKahane:
Git error, this branch is not supposed to have all the commits of main :/
I will create a new branch

EdenKahane:
Close due to git problem

PR #84, opened 2024-10-11
Code restructuring and many small fixes

EdenKahane:
This PR helps with many small fixes that were needed through the code but might not worth one PR by themselves.

- Code restructuring to follow app architecture
- File moved to follow the new app architecture
- function now use Result when needed
- Updated README with team minutes and team review links
- Fixed README indentation issue
- Moved ColorVariables in an object for better code understanding
- Fixed gitignore not excluding  app/google-services.json
- Fixed variable name convention issues
- Fixed wrong comments used for function description

The actual code and tests content of some of these files, need to be recoded in a separate task (already created). This PR is just for the code structure and minor fixes.


app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
```diff
@@ -43,24 +48,24 @@ class Geolocation(private val activity: Activity) {
   private val locationCallback =
       object : LocationCallback() {
         override fun onLocationResult(p0: LocationResult) {
-          p0.lastLocation?.let { location ->
+          p0.lastLocation.let { location ->
```
TheoSchlaeppi, line 51: We concluded in a previous PR that it would be better to check lastLocation, even if the IDE says it's unnecessary.

TheoSchlaeppi (approved):
Except for the lastLocation case, all the changes were welcome and necessary for the structure of the code. Thank you for this PR. I approve the code, but you need to either prove that the check is indeed unnecessary or make a commit to restore the '?'.

app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
EdenKahane: LocationCallback specify the `@NonNull` annotation denoting that the parameter can never be null. We can thus safely remove the '?'

```kotlin
public abstract class LocationCallback {
    public LocationCallback() {
    }

    public void onLocationResult(@NonNull LocationResult var1) {
    }

    public void onLocationAvailability(@NonNull LocationAvailability var1) {
    }
}
```

PR #85, opened 2024-10-13
feat: Add notification actions for Accept and Decline in FirebaseMess…

matiisalv:
…agingService and handle notification action

- Implemented PendingIntent for "Accept" and "Decline" actions in notifications.
- Created NotificationActionReceiver to handle action intents.
- Updated notification channel creation for Android O and above.
- Implemented `NotificationActionReceiver` class extending `BroadcastReceiver`
- Added `onReceive` method to handle "ACTION_ACCEPT" and "ACTION_DECLINE" intents
- Adding function to ask and require permission, but full incorporate in the UI part of the notification
-Added and firebase message service which send notifications when a message is send


app/src/main/java/com/android/bookswap/model/chat/FirebaseMessagingService.kt
```diff
@@ -0,0 +1,113 @@
...
+/** Service to handle Firebase Cloud Messaging notifications. */
+open class MyFirebaseMessagingService : FirebaseMessagingService() {
+  /**
+   * Called when a message is received.
+   *
+   * @param remoteMessage the message received from Firebase Cloud Messaging.
+   */
+  override fun onMessageReceived(remoteMessage: RemoteMessage) {
+    val notification = remoteMessage.notification
+    val title = notification?.title ?: "Hello There!"
```
TheoSchlaeppi, line 23: This is hard-coded. Add a docstring to remind us to update it later when we have the string file.

app/src/main/java/com/android/bookswap/model/chat/FirebaseMessagingService.kt
```diff
@@ -0,0 +1,113 @@
...
+   * Sends a notification to the user.
+   *
+   * @param title the title of the notification.
+   * @param messageBody the body of the notification.
+   */
+  fun sendNotification(title: String, messageBody: String) {
+    // Create an intent to open the MainActivity when the notification is clicked
+    val intent =
+        Intent(this, MainActivity::class.java).apply { addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP) }
+    val pendingIntent =
```
TheoSchlaeppi, line 38: The request code is the same for both 'accept' and 'decline' (request code = 0). It would be better to use two different request codes to avoid potential conflicts. You can refer to this [StackOverflow thread](https://stackoverflow.com/questions/21526319/whats-requestcode-used-for-on-pendingintent) for more information on the issues that may arise.

app/src/main/java/com/android/bookswap/model/chat/FirebaseMessagingService.kt
```diff
@@ -0,0 +1,113 @@
...
+        Intent(this, NotificationActionReceiver::class.java).apply { action = "ACTION_DECLINE" }
+    val declinePendingIntent =
+        PendingIntent.getBroadcast(
+            this,
+            0,
+            declineIntent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
+
+    // Define a meaningful channel ID and channel name
+    val channelId = "bookswap_notifications_channel"
```
TheoSchlaeppi, line 63: Also add docstring here to notify the hard coding

app/src/main/java/com/android/bookswap/model/chat/FirebaseMessagingService.kt
TheoSchlaeppi: And you can also create Constant variable for the number choose Ex: REQUEST_CODE_ACCEPT

app/src/main/java/com/android/bookswap/model/chat/FirebaseMessagingService.kt
```diff
@@ -0,0 +1,113 @@
...
+  fun createNotificationChannel(
+      notificationManager: NotificationManager,
+      channelId: String,
+      channelName: String
+  ) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      val importance = NotificationManager.IMPORTANCE_HIGH
+      val channel =
+          NotificationChannel(channelId, channelName, importance).apply {
+            description = "Channel for BookSwap notifications"
```
TheoSchlaeppi, line 105: I don't know if this count as hard coded but if you think so, add also docstring

app/src/test/java/com/android/bookswap/model/chat/MyFirestoreMessagingServiceTest.kt
```diff
@@ -0,0 +1,58 @@
...
+            service.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager)
+
+    // Assert that a notification was posted
+    val postedNotifications = notificationManager.allNotifications
+    assertNotNull("Notification should have been posted", postedNotifications)
+    assertTrue("Notification list should not be empty", postedNotifications.isNotEmpty())
+  }
+
+  @Test
+  fun testOnMessageReceived_withNotificationPayload() {
```
TheoSchlaeppi, line 46: For this kind of test i think that you can use convention that we have on Discord.

TheoSchlaeppi (changes requested):
Well done on this implementation! The code is well-structured and covers the main functionalities. I have a few suggestions for improvement:

### 1. Hardcoded Values
- There are some hardcoded strings, such as the notification titles and messages. It would be a good idea to adding a docstring could to facilite the creation of strings.xml.

### 2. Request Code for PendingIntent
- The request code is the same for both the "Accept" and "Decline" actions (`0`). Using different request codes would help avoid potential conflicts. You can find more details about this on the link shared.


### 3. Testing
- The tests look solid! You could also add checks for the content of the notifications (e.g., title, message) to ensure they are set correctly.

Overall, great work! These suggestions are just minor tweaks that could enhance the code further.

EdenKahane (changes requested):
Could you explain why not restricting the service in AndroidManifest is not a problem ? 
It's often used as a vector exploit in apps even if there is a client check.
There is some solutions [here](https://github.com/BookswapEPFL/Bookswap/security/code-scanning/1)

matiisalv:
I was thinking implementing the restrictions in the UI  when it asking for permission to limit it but I can do it in this part if you consider more appropriate

TheoSchlaeppi (approved):
Looks good for me. Good job !

EdenKahane (approved):
Seems good.
Very well tested too, especially for toasts !

PR #87, opened 2024-10-15
Adds tests for the UI of the chat

oliveja22:
Includes all tests for the UI of the chat, including everything being displayed when no messages are there. Mocking sending 6 messages to test all components of messages, including text, bubble and timestamp. Includes testing that textField works and button is clickable. Includes testing that button sends message by mocking sending a message and generating and ID. Also includes a small test for the function that creates the timestamps.

Also added a test in DataMessageFirestoreSourceTest for documentToMessage function

Includes modifications to the ChatScreen found during testing, including testTags and null checks.


LeniCadorinEpfl (changes requested):
Great job on writing these complete tests! Aside from the CI issue, the test code looks great. Once the CI problem is resolved, it's good to go for merging from my side.

LeniCadorinEpfl (approved):

PR #89, opened 2024-10-16
Feature: Implementation of the filter screen

TheoSchlaeppi:
This PR introduces the `FilterMapScreen` and `ButtonBlock` composables, which allow users to select filters for book genres and languages. The `FilterMapScreen` provides a UI for filtering, including a top bar, a content area with buttons for genres and languages, and a bottom bar with an "Apply" button.

- Added `FilterMapScreen` composable:
  - Displays a top bar with a title and back button.
  - Contains a `LazyColumn` for displaying selectable filter buttons (genres and languages).
  - Includes a bottom bar with an "Apply" button that adds the selected filters and navigates back.
- Added `ButtonBlock` composable:
  - Uses `FlowRow` to display a row of buttons.
  - Buttons toggle their selection state and update the list of selected filters.
- Added a new `Genre` enum class to represent book genres, making the code more type-safe and readable.
  - Introduced two new `const val` lists:
    - `LIST_BOOK_GENRES`: Contains all available genres.
    - `LIST_BOOK_LANGUAGES`: Contains all available languages.

image

image

## Additional Notes

- Some hardcoded strings and styles should eventually be extracted to string and theme files.
- The integration of this screen into the map screen will be handled in the upcoming task focused on filter logic implementation.
- The Genre enum class needs further refinement and should be integrated into the Book data class, which will be addressed in a separate task.


app/src/main/java/com/android/bookswap/data/DataBook.kt
```diff
@@ -35,3 +35,58 @@ enum class BookLanguages(val languageCode: String) {
   ROMANSH("RM"), // Romansh, a language spoken in Switzerland
   OTHER("OTHER") // All languages that are not yet implemented
 }
+
+/** List of all supported book languages */
+val LIST_BOOK_LANGUAGES =
+    listOf(
```
Chockencore, line 41: It isn't a good practice to do the list manually as if in the future we need to add a new language we may forget to add to this list. This list can be remplaced by BookLanguages::class.java.enumConstants?.map { it.name }. Therefore I think this value can be removed.

app/src/main/java/com/android/bookswap/ui/map/filterMap.kt
```diff
@@ -0,0 +1,178 @@
...
+fun ButtonBlock(buttonTexts: List<String>, selectedFilters: MutableList<Any>) {
+  FlowRow(
+      modifier =
+          Modifier.fillMaxWidth()
+              .padding(
+                  top = VERTICAL_PADDING_BB,
+                  start = HORIZONTAL_PADDING_BB,
+                  end = HORIZONTAL_PADDING_BB,
+                  bottom = VERTICAL_PADDING_BB),
+      maxItemsInEachRow = 3) {
```
Chockencore, line 145: The 3 should be a constant

Chockencore (changes requested):
The code is well done, there is just some details to change.
1) The name of the file should always begins by an uppercase letter. It should be FilterMap.
2)The LIST_BOOK_LANGUAGES isn't necessary as the it can be remplaced in the code by BookLanguages::class.java.enumConstants?.map { it.name }
3)In the composable ButtonBlock, there is just one hard coded value that wasn't put in the constants.
4)It's strange to have the filter list in the argument. It should be stored in a viewModel as it is used in multiple screens and the logic should be also in this viewModel even if the logic takes only a few lines of code. If you consider that creating the new viewModel should be another task because you consider it would too much time or for another reason, the list of filters can stay in the arguments but the comments should specify that it is only temporary until someone creates this new viewModel.

LeniCadorinEpfl (commented):
I just saw what Sebastien said, and I totally agree. I'll just add a few more suggestions:

1. Some tests could be added for edge cases, such as ensuring the correct behavior when no filters are selected.
2. A toast could be displayed when the filter has been correctly applied, so the user is notified of the changes.
3. Finally, just don't forget to add the screen to the navigation.

Overall, I'd say the code is great. These are just points to make it even better. Also if any of the comments are unclear just let me know.

PR #90, opened 2024-10-16
Add workflow to build an APK

Chockencore:
This PR introduces an new workflow called Build APK that creates an APK of the app. It can be triggered manually from the github UI and is automatically triggered by pushes on main, M1, M2, M3, apkBuild. The M1, M2 and M3 branches should be only used for the milestones and the apkBuild branch can be used to create the APK for the end of each sprint. The workflow uploads the APK in the artefact of the workflow.


Chockencore:
After many tries I couldn't make the google signIn work with the APK from the workflow, but the rest of the app should work fine. It should be added in a future sprint, but there is normally a good base to make it work.

Issue #81, opened 2024-10-11
Refactor theme and color useage to use theme.kt and color.kt


Chocolat2:
# Refactore the entire codebase to ensure the use of the Bookswap app theme and colors everywhere in the app

This issue is composed of multiple tasks:

- [x] #158
- [x] #159
- [x] #160
- [x] Remove all the theme, color and style overrides in the code (f.e. `Modifier.background(ColorVariable.BackGround)`)
- [x] Revise existing custom components:
  - [x] Ensure [Figma's design](https://www.figma.com/design/uyHS0PV5RBnmToK2JuTb4h/Bookswap?node-id=0-1&t=nQQFtO8r6GVH1yCR-1) is properly implemented
  - [x] #161
- [x] Create custom components like the ButtonComponent and FieldComponent to customize default Material3 components when necessary

PR #92, opened 2024-10-17
Feature: Implementation of the bottomNavigationMenu

LeniCadorinEpfl:
This PR introduce the `bottomNavigationMenu` composable, wich allows to navigate between the different screen. It also includes test cases for verifying the behavior of the Bottom Navigation Menu in the UI.

Bottom Navigation Menu composable:
Displays the bottom navigation bar using the `BottomNavigation` composable.
Each tab is represented by an icon.

Screenshot 2024-10-17 005422

**Checklist:**
- [x] CI passes
- [x] Structure is correct


app/src/main/java/com/android/bookswap/ui/navigation/BottomNavigationMenu.kt
```diff
@@ -0,0 +1,46 @@
...
+) {
+  BottomNavigation(
+      modifier = Modifier.fillMaxWidth().height(BOTTOM_NAV_HEIGHT).testTag("bottomNavigationMenu"),
+      backgroundColor = ColorVariable.Primary, // Color of the bottom navigation bar
+      content = {
+        tabList.forEach { tab ->
+          BottomNavigationItem(
+              icon = {
+                // Display the icon for each tab
+                Icon(tab.icon, contentDescription = null, tint = ColorVariable.BackGround)
```
TheoSchlaeppi, line 38: Wouldn't it be better if the contentDescription was set to tab.textId?

TheoSchlaeppi (approved):
The code is well-structured, and the tests cover essential functionalities. Here are a suggestions to improve the code and enhance accessibility.

Currently, the `contentDescription` is set to `null`. To improve accessibility, it would be better to set it to `tab.textId`, which would allow screen readers to announce the name of the tab. For example:

```kotlin
Icon(tab.icon, contentDescription = tab.textId, tint = ColorVariable.BackGround)
```

Thank you for this necessary component. I approve and as it's a very small change i will modify it for on your branch before merging it.

app/src/main/java/com/android/bookswap/ui/navigation/BottomNavigationMenu.kt
TheoSchlaeppi: Finally i change the code with 
```kotlin
Icon(tab.icon, contentDescription = "icon_${tab.textId}", tint = ColorVariable.BackGround)
```
To avoid conflict with testTag

app/src/main/java/com/android/bookswap/ui/navigation/BottomNavigationMenu.kt
```diff
@@ -0,0 +1,49 @@
...
+              icon = {
+                // Display the icon for each tab
+                Icon(
+                    tab.icon,
+                    contentDescription = "icon_${tab.textId}",
+                    tint = ColorVariable.BackGround)
+              },
+              selected = tab.route == selectedItem,
+              onClick = { onTabSelect(tab) },
+              modifier = Modifier.clip(RoundedCornerShape(ROUNDED_CORNER_SIZE)).testTag(tab.textId))
```
Chocolat2, line 45: Is there a specific reason why you added the clipping on each icon ?
If it is to get a visual effect on the tabs, the `NavigationBarItem`s already display as a rounded-corner button when selected, focused or hovered so it is not necessary. (and can lead to improper display of the icons)

Chocolat2 (commented):
Code looks good overall, but some changes will be needed in the future.

The navigation actions should be added as soon as possible.
The selected state of each tab should be fixed.
The component specific color and style should be removed so that the custom Material theme is used across the app.
The composable should be updated to use Material3 components as it will cause problems when applying the custom theme `BookSwapAppTheme`.

Some of the tests can be expanded to cover a longer flow of actions such as navigating multiple times to the same tab or navigating to multiple tabs and ensuring everything holds well at each step.

Chocolat2:
> The navigation actions should be added as soon as possible. The selected state of each tab should be fixed. 
After further discussing how the `NavigationBar` is inserted in the app, the navigation action will be handled using the callback by the parent component and the selected state will be set to the proper value at each recomposition of the `BottomNavigationMenu` component.

> The component specific color and style should be removed so that the custom Material theme is used across the app.
This refactoring is planned in issue #81 and will be done in a future pull request.

> The composable should be updated to use Material3 components as it will cause problems when applying the custom theme `BookSwapAppTheme`.
Can be done while refactoring the color and style in #81.

> Some of the tests can be expanded to cover a longer flow of actions such as navigating multiple times to the same tab or navigating to multiple tabs and ensuring everything holds well at each step.
Not planned but can be done later on.

PR #94, opened 2024-10-17
feature: Create UI screen to create a book from ISBN

EdenKahane:
Add Screen UI for the page that let the user add a book based on a isbn.
This PR use methods of https://github.com/BookswapEPFL/Bookswap/pull/50 and add configurable UI components that can be used in other pages

- [x] Add tests

This PR is a git fix of #74


Chockencore (changes requested):
The code is clean but there are some problem.
1) The hard coded value should be constants

2)There is already a back button created in the navigation folder, the screen should use it

3)The search button crashes the app when entering a wrong ISBN, it should distinguish between a failure in the code and a wrong ISBN.

Chockencore (approved):
Looks good for me, there is a need to put the hard coded values in constants bit that can be done in another task

PR #97, opened 2024-10-17
Feature/chocolat2/user repo

Chocolat2:
Implement the data source for user and add a Firebase data source for user data


app/src/main/java/com/android/bookswap/ui/profile/EditProfile.kt
```diff
@@ -0,0 +1,137 @@
...
+/*@Preview(showBackground = true, widthDp = 540, heightDp = 1110)
+@Composable
+fun EditProfileDialoguePreview() {
+  val address = android.location.Address(java.util.Locale.getDefault())
+  address.countryCode = "CH"
+  address.locality = "Lausanne"
+  address.postalCode = "1000"
+  address.countryName = "Switzerland"
+  address.setAddressLine(0, "Rue de la Gare 1")
+  EditProfileDialog(onDismiss = { *//*TODO*//* }, onSave = { *//*TODO*//* }, dataUser = DataUser("Mr.", "John", "Doe", "John.Dow@example.com","+41223456789",address,"dummyPic.png","dummyUUID0000"))
```
Chockencore: Why is this code in comment? If it isn't supposed to be used, then it needs to be deleted.

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
```diff
@@ -0,0 +1,168 @@
...
+              // Title of the screen
+              title = { Text("Your Profile", modifier = Modifier.testTag("profileTitleTxt")) },
+              // Icon button for navigation (currently no action defined)
+              navigationIcon = {
+                IconButton(modifier = Modifier.size(32.dp), onClick = {}) {
+                  Icon(
+                      imageVector = Icons.AutoMirrored.Sharp.ArrowBack,
+                      contentDescription = "",
+                      modifier = Modifier.size(30.dp))
+                }
```
Chockencore, line 90: Should use the already created back button.

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
```diff
@@ -0,0 +1,168 @@
...
+                            Image(
+                                imageVector = Icons.Rounded.AccountCircle,
+                                contentDescription = "",
+                                modifier = Modifier.fillMaxSize().scale(1.2f).clipToBounds(),
+                                colorFilter = ColorFilter.tint(Color(0xFF6C584C)))
+                          }
+                      Box(
+                          modifier = Modifier.fillMaxSize().padding(0f.dp),
+                          contentAlignment = Alignment.TopEnd) {
+                            IconButton(onClick = { /*TODO*/}, modifier = Modifier) {
```
Chockencore, line 114: You should put a description in comment when you are using a TODO, so we know in the future what you wanted.

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
```diff
@@ -0,0 +1,168 @@
...
+//      "Doe",
+//      "John.Doe@example.com",
+//      "+41223456789",
+//      0.0,
+//      0.0,
+//      "dummyPic.png",
+//      "dummyUUID0000")
+//  )
+//  UserProfile(userVM)
+// }
```
Chockencore, line 173: Need to remove the commented code

app/src/main/java/com/android/bookswap/ui/profile/EditProfile.kt
```diff
@@ -0,0 +1,137 @@
...
+import androidx.compose.ui.window.DialogProperties
+import com.android.bookswap.data.DataUser
+import com.android.bookswap.ui.theme.BookSwapAppTheme
+
+@Composable
+fun EditProfileDialog(
+    onDismiss: () -> Unit,
+    onSave: (DataUser) -> Unit,
+    dataUser: DataUser,
+    modifier: Modifier = Modifier
```
Chockencore: The parameter modifier is never used, if it isn't supposed to be used, then it can be removed.

app/src/main/java/com/android/bookswap/ui/profile/EditProfile.kt
```diff
@@ -0,0 +1,137 @@
...
+) {
+
+  var _email = remember { mutableStateOf<String>(dataUser.email) }
+  var _phone = remember { mutableStateOf<String>(dataUser.phoneNumber) }
+  var _greeting = remember { mutableStateOf<String>(dataUser.greeting) }
+  var _firstName = remember { mutableStateOf<String>(dataUser.firstName) }
+  var _lastName = remember { mutableStateOf<String>(dataUser.lastName) }
```
Chockencore: All of them can be declared as val

Chockencore (changes requested):
The code this PR introduces is good, there are just details to fix.
There are some codes that was put as comments, they need to be deleted, one parameter in EditProfileDialog isn't used, the variables at the beginning of EditProfileDialog  can be changed to val.
The file profileIcon should have a upper case letter so it should be ProfileIcon and should be put in the components folder (the main needs to be merge into the brach before as it doesn't yet exist)
Also, the screens sould be added in the navhost and Navigation file as screen, and there should be a bottom navigation bar but this part can be addressed in another PR as it would be too much for this already big PR.

app/src/main/java/com/android/bookswap/ui/profile/EditProfile.kt
Chocolat2: This is a preview used in Android Studio to make UI design easier and faster by being able to render the UI without launching the app on the emulator, however , there is no way to "test" this part so it is currently commented to avoid reducing line coverage

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
Chocolat2: This is a preview used in Android Studio to make UI design easier and faster by being able to render the UI without launching the app on the emulator, however , there is no way to "test" this part so it is currently commented to avoid reducing line coverage

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
Chocolat2: This is a preview used in Android Studio to make UI design easier and faster by being able to render the UI without launching the app on the emulator, however , there is no way to "test" this part so it is currently commented to avoid reducing line coverage

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
Chocolat2: Not really relevant to this PR since the PR is for the data accesses using repository and source. It will be done in a future PR.

app/src/main/java/com/android/bookswap/ui/profile/EditProfile.kt
Chocolat2: The modifier is used when this composable is instantiated as part of another component, just like the modifiers from basic Material3 components

Chockencore (approved):
It looks good to me

PR #101, opened 2024-10-17
Implement the logic behind the filter of the books on the map screen

TheoSchlaeppi:
This PR introduces the logic for filtering books on the `MAP` route.

### Book Filter
The model shared by the Map and Filter screens contains two StateFlows for languages and genres, which the filter will apply to the list of books on the map. This model required a small modification to the Map screen, including the creation of new variables `filteredBooks` and `filteredUsers`, allowing the original lists of books and users to remain unchanged. `filteredBooks` is now the new parameter for the DraggableMenu, and the placement of markers for each user has been adjusted to place markers for each `filteredUser`.

This PR also includes tests related to this model. The logic behind filtering by language and genre is not the same, as the enum classes differ.

### MainActivity
In this PR, the NavHost has been modified to add the creation of the Book Filter, which is passed as a parameter to both the Filter screen and the Map screen.

All suggestions are welcome, and I am available on this thread for any questions.


app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
```diff
@@ -32,7 +36,8 @@ class MapScreenTest {
                           rating = 5,
                           photo = "url_to_photo_1",
                           language = BookLanguages.ENGLISH,
-                          isbn = "123-456-789"),
+                          isbn = "123-456-789",
+                          genres = listOf(BookGenres.FICTION, BookGenres.NONFICTION)),
```
oliveja22, line 40: Nice combination of genres

app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
```diff
@@ -130,7 +161,11 @@ class MapScreenTest {
 
   @Test
   fun draggableMenu_canBeDraggedVertically() {
-    composeTestRule.setContent { MapScreen(user, user[0]) }
```
oliveja22, line 133: The title of the test is a bit inconsistent here, you seem to use camelCase everywhere but one word in this test title. Though I must say that given everything that could either change or have gone wrong, this one is practically insignificant.

app/src/main/java/com/android/bookswap/MainActivity.kt
```diff
@@ -103,4 +112,6 @@ val user =
                         rating = null,
                         photo = null,
                         language = BookLanguages.GERMAN,
-                        isbn = null))))
+                        isbn = null,
```
oliveja22, line 115: Nice mix of genres again.

app/src/main/java/com/android/bookswap/model/map/BookFilter.kt
```diff
@@ -0,0 +1,53 @@
+package com.android.bookswap.model.map
+
```
oliveja22, line 2: I do have to say, I find it somewhat difficult to ensure my comprehension of the functions below without some aid from comments. It would certainly be well welcome to see some comments added for each function below.

app/src/main/java/com/android/bookswap/ui/map/FilterMap.kt
```diff
@@ -89,18 +90,22 @@ fun FilterMapScreen(navigationActions: NavigationActions, selectedFilters: Mutab
         LazyColumn(contentPadding = paddingValues, modifier = Modifier.fillMaxSize()) {
           item {
             ButtonBlock(
-                BookGenres.values().map {
-                  it.toString().lowercase().replaceFirstChar { c -> c.uppercase() }
-                },
-                selectedFiltersTemp)
+                BookGenres.values().map { it.Genre }, selectedFiltersGenres.map { it.Genre }) {
```
oliveja22, line 93: I seem to miss some alignment modifiers, maybe some padding. Through this I cannot sight what the final screen may look like, but it seems like some standardised padding and some central alignment may benefit. That being said, and as you may with great ease realise, it may look quite alright as of now. If that is the case, please ignore this message.

app/src/main/java/com/android/bookswap/ui/map/FilterMap.kt
```diff
@@ -89,18 +90,22 @@ fun FilterMapScreen(navigationActions: NavigationActions, selectedFilters: Mutab
...
+                    newSelection ->
+                  bookFilter.setGenres(newSelection)
+                }
           }
           item {
```
oliveja22, line 98: It would be nice to have some comments here as well, I find myself reading things anew more than I would normally wish, and while it seems to be fine, it does show that perhaps some comments might aid in its lecture.

oliveja22 (approved):
I believe the work you have here shown is quite marvelous. There seems to be a distinct lack of comments; the code would certainly find great benefit in said comments, especially when it comes to its readability. Otherwise one may discover a few naming inconsistencies, however they are as minor as problems may reach. Given this, I have not much to say other than praise for the code. It seems to work quite well, and I find it quite interesting. I find particularly interesting how you filter the books, which is done cleverly.
Perhaps one last detail to mention is, that some parts of the user interface lack modifiers, this may not have any effect as I said in my comments in the code, but it may be worth a quick look at. Additionally, I find that one test might be missing, as we both know, is that the buttons change colour when clicked. This may be worth having a look at. Now, other than what I just said, and like I said at the very beginning, I find your code marvelous, and somewhat fascinating, especially in how you filter the books.

app/src/main/java/com/android/bookswap/model/map/BookFilter.kt
TheoSchlaeppi: Yeah I forgot to add docs but i just commited it right now

app/src/main/java/com/android/bookswap/ui/map/FilterMap.kt
TheoSchlaeppi: Fixed  in the FlowRow

PR #102, opened 2024-10-17
fix:This PR fix the problem of  Bookrepositorywith the new fileds Boo…

matiisalv:
…kGenres and change the println of AddToBook into Log.e

This PR the issue with BookRepository and new the field


app/src/main/java/com/android/bookswap/model/chat/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,118 @@
...
+            acceptIntent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
+
+    // Create intent and PendingIntent for the "Decline" action
+    val declineIntent =
+        Intent(this, NotificationActionReceiver::class.java).apply { action = "ACTION_DECLINE" }
+    val declinePendingIntent =
+        PendingIntent.getBroadcast(
+            this,
+            0,
```
LeniCadorinEpfl, line 62: Both the "Accept" and "Decline" actions use the same request code (0) for creating the PendingIntent. This could cause unexpected behavior since they are distinct actions but share the same request code.

app/src/main/java/com/android/bookswap/model/chat/NotificationActionReceiver.kt
```diff
@@ -0,0 +1,33 @@
...
+        requestCode(context, "Accepted")
+      }
+      "ACTION_DECLINE" -> {
+        // Handle decline action
+        requestCode(context, "Declined")
+      }
+    }
+  }
+
+  private fun requestCode(context: Context, action: String) {
```
LeniCadorinEpfl, line 29: I think the naming is a bit strange, it is misleading as it doesn’t return or handle request codes but shows a Toast.
The name could be changed to enhance understanding of the code.

LeniCadorinEpfl (approved):
The code is well structured, the comment in the code are a good practice and are really useful. Here are a just suggestions to improve the code : 

1. Renaming: 
The method name `requestCode` is a bit misleading as it doesn’t return or handle request codes but shows a Toast.
Suggestion: Rename requestCode to something more descriptive like handleAction, which better represents its functionality.
2. Request code :
Both the "Accept" and "Decline" actions use the same request code (0) for creating the PendingIntent. This could cause unexpected behavior since they are distinct actions but share the same request code.
Suggestion: Use unique request codes for each action to avoid conflicts:
3. Test: Unit test could be added for the new feature

app/src/main/java/com/android/bookswap/model/chat/MyFirebaseMessagingService.kt
matiisalv: yes it 's fix in  an  another PR , but thanks you for the suggestion

PR #107, opened 2024-10-17
Fix Listener for Chat

oliveja22:
This pull request seeks to temporarily fix the issue with the listener.


EdenKahane (changes requested):
A working temp fix, will be working on it again later

EdenKahane (approved):

PR #117, opened 2024-10-17
Implement the global navigation for MS1

LeniCadorinEpfl:
This PR introduce the global navigation for the MS1. It allows to navigate through the different major screens already implemented.  

#### Changes:

1. Update the `MainActivity`  to include the navigation to each Screen
2. Update test to include the navigation
3. Update each screen signature to include the navigation
4. Include the ` BottomNavigationMenu` in the screen that needed it:  `MapScreen`, `listChatScreen`, `BookAdditionChoiceScreen`, 
5. Include the `BackButton` in the screen that need it:  `listChatScreen`, `AddToBookScreen`.

**Checklist:**
- [x] CI passes
- [x] Structure is correct


Chockencore (changes requested):
The code is clean but the Add ISBN Screen was not added in the navhost in the main activity and there is no back button for the chat screen. Otherwise everything is good.

LeniCadorinEpfl:
> The code is clean but the Add ISBN Screen was not added in the navhost in the main activity and there is no back button for the chat screen. Otherwise everything is good.

Thank you for your review, you're absolutely right. Regarding the ISBN screen, it was added just a few minutes before this PR, which is why it wasn't included. We can definitely address it in a future task.

As for the back button in the ChatScreen, after discussing with Jaime, we agreed that implementing the back button would be challenging due to the current structure of his Compose code, which doesn't use a Scaffold and is a bit different from the standard setup. We've decided to handle this in a separate task to ensure it’s done properly.

Chockencore (approved):
The screen to add a book with the ISBN was added, so it's good for me, the back button is too much for this sprint.

PR #118, opened 2024-10-18
Fix the draggable menu as it was going behind the bottom navigation bar

Chockencore:
This PR fixes the bug where the draggable menu was disappearing behind the navigation bar. The lowest position of the draggable menu was raised to compensate the height of the bottom navigation bar.


LeniCadorinEpfl (approved):
Needed change, great work!

PR #119, opened 2024-10-18
Implement end 2 end test for navigation through top level destination

TheoSchlaeppi:
This PR introduces a new test that navigates through all top-level destinations already connected to the NavHost. It will be necessary to add the path to the profile route once this route is linked in the NavHost. I also modified the signature of BookSwapApp to allow selecting the start destination.


Chockencore (changes requested):
The test is good but it starts at the map screen and the first thing it tries to do is to go to the map screen, which doesn't confirm if the button really did anything. This part should be put after the add screen or the chat screen.

Chockencore (approved):
Looks good to me

PR #120, opened 2024-10-18
Feature add toasts to signal informations to user

EdenKahane:
Add toasts to various screens


EdenKahane:
This solve #130

TheoSchlaeppi (approved):
Thank you for these changes, which significantly enhance the user experience with the toast notifications when actions occur, as well as improve code maintainability for developers. Regarding the test coverage, all uncovered lines appear to be placeholders, so it seems there isn't much more that can be done there. Overall, great job!

EdenKahane:
Thank you for the review ! 
There is more that can be done, but it's from a previous code part and adding tests for that is out of the scope of this task.
Now with mockk it shouldn't be too hard :D

PR #133, opened 2024-10-29
Fix: make the draggable menu handle the scroll properly

Chockencore:
This PR fixes the problem that the composable "DraggableMenu" had when trying to scroll the list of books inside it. With the current list of books put as an example in the app and with the default position of the menu, it appeared that it wasn't possible to scroll the list of books. 
It wasn't the case, the problem was that the list continued to be displayed behind the navigation bar, so for the app, the list was already fully displayed. Therefore, it wasn't possible to scroll.  The offset and the height limits of the draggable menu were modified to take into account the height of the navigation bar.


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -311,108 +308,104 @@ private fun DraggableMenu(listAllBooks: List<DataBook>) {
...
+                      .testTag("mapDraggableMenuHandle"))
+          Spacer(modifier = Modifier.height(SPACER_HEIGHT_DP.dp))
+          HorizontalDivider(
+              modifier = Modifier.fillMaxWidth().testTag("mapDraggableMenuHandleDivider"),
+              thickness = DIVIDER_THICKNESS_DP.dp,
+              color = ColorVariable.Accent)
+          LazyColumn(userScrollEnabled = true, modifier = Modifier.fillMaxHeight()) {
+            itemsIndexed(listAllBooks) { index, book ->
+              Spacer(modifier = Modifier.height(PADDING_VERTICAL_DP.dp))
+              Row(
```
oliveja22, line 334: You should probably add an onClick(), it might be empty right now, but we will need it at some point, it will redirect to it's profile normally.

oliveja22 (approved):
The code in itself is quite solid, other than adding an onClick() to the row where the comment has been added, there is not much to say. There are enough tests and the placeholders are quite well implemented to be easily replaced. Nice job

PR #134, opened 2024-10-29
Complete Book Profile Page added

oliveja22:
Includes a book profile screen that takes in a DataBook and navigation Actions to display all information as necessary in an organised manner. Also includes two placeholder images used for testing as the DataBook does not have the photo implemented yet.
Also includes a large number of placeholders for the moment, as they are not implemented into DataBook and may be removed at a later fix.


EdenKahane (approved):
Perfect PR.
The UI tests works well and use the new system.
So all good.

PR #136, opened 2024-10-30
Expand Messaging Function to include deleting and editing

oliveja22:
Includes new UI parts for editing and deleting, a pop-up upon a long press of a message with two buttons delete and edit, which can be pressed to perform their functions. Delete simply deletes the text and update is more complex, redirects to the text field which is filled with old message, it can be modified, and send button changed to update button, when pressing it updates the text and timestamp. 
Also includes the firebase implementation of said functions, which also include the caveat that they can only be edited or deleted if less than 15 minutes have passed. Upon failing to meet this condition, nothing will happen and a toast will be shown.
It also added a top app bar to the chat and other small details.
Also includes tests for all of this, 3 new tests in the androidTest, including one to check that the pop up appears, one to check that deleting works, and one for editing (the most complex one). There are additionally numerous tests for multiple cases in the unitTests, including adding some for some cases for previous functions that were missing some tests.

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/ui/chat/ChatScreenTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/ui/chat/ChatScreenTest.kt
@@ -187,6 +187,22 @@ class ChatScreenTest {
       callback(sendMessageResult)
     }
 
+    override fun deleteMessage(message: DataMessage, callback: (Result<Unit>) -> Unit) {
+      TODO("Not yet implemented")
+    }
+
+    override fun deleteAllMessages(
+        user1Id: String,
+        user2Id: String,
+        callback: (Result<Unit>) -> Unit
+    ) {
+      TODO("Not yet implemented")
+    }
+
+    override fun updateMessage(message: DataMessage, callback: (Result<Unit>) -> Unit) {
+      TODO("Not yet implemented")
+    }
+
     override fun addMessagesListener(
         otherUserId: String,
         currentUserId: String,
--- a/app/src/androidTest/java/com/android/bookswap/ui/chat/ChatScreenTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/ui/chat/ChatScreenTest.kt
@@ -1,14 +1,22 @@
 package com.android.bookswap.ui.chat
 
+import android.content.Context
+import androidx.compose.ui.semantics.SemanticsActions
+import androidx.compose.ui.semantics.SemanticsProperties
+import androidx.compose.ui.semantics.getOrNull
 import androidx.compose.ui.test.assertHasClickAction
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertTextEquals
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
```diff
@@ -56,6 +58,141 @@ class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageReposit
...
+            } else {
+              callback(Result.success(Unit))
+            }
+          } else {
+            callback(Result.failure(task.exception ?: Exception("Unknown error fetching messages")))
+          }
+        }
+  }
+
+  override fun updateMessage(
```
matiisalv, line 142: Maybe it would better to use the name of EditMessage

app/src/androidTest/java/com/android/bookswap/ui/chat/ChatScreenTest.kt
```diff
@@ -187,6 +384,39 @@ class ChatScreenTest {
...
+      callback(Result.success(Unit))
+    }
+
+    override fun updateMessage(
+        message: DataMessage,
+        callback: (Result<Unit>) -> Unit,
+        context: Context
+    ) {
+      val index = messages.indexOfFirst { it.id == message.id }
+      if (index != -1) {
```
matiisalv, line 412: Could use a constant out of band more than a number for clarity

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
```diff
@@ -67,70 +85,184 @@ fun ChatScreen(
...
+            }
+      }
     }
   }
 }
 
+@OptIn(ExperimentalFoundationApi::class)
 @Composable
-fun MessageItem(message: DataMessage, currentUserId: String) {
+fun MessageItem(message: DataMessage, currentUserId: String, onLongPress: () -> Unit) {
```
matiisalv, line 265: Could be add   a second argument onClick , because it possible that the action  to click is different if it's long or short could be different

matiisalv (changes requested):
It's a very nice feature that you added here! You did a wonderful job; it’s well-written and well-documented. However, I have some recommendations that could help you.

First of all, please add more tests, for example, for successful delete and edit actions. I think this could give you the necessary coverage and won't be that difficult to test. Secondly, be careful with the MessageItem; I wrote a specific comment directly about it.

Other than that, your work is excellent. Good job

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
oliveja22: There is just no action for a click, so it makes no sense to add it, if I did, it would be left empty anyway, additionally, a OnLongPress does not trigger with a simple click, so adding it is unnecessary

matiisalv (approved):
Thanks for adding the extra tests, more are needed but we can create a new task for it as we discussed.

PR #137, opened 2024-10-30
Small Fix of test for the list of chat

TheoSchlaeppi:
This PR addresses an issue with tests that previously couldn’t verify behavior when there were no messages, ensuring accurate checks in those cases. It also adds a validation for the navigation bar on this screen to enhance UI consistency and test reliability.


oliveja22 (approved):
Really nothing to say here, just a few lines of code for the tests, if they pass, which they seemingly do, I have nothing to say, good or bad.

PR #144, opened 2024-10-31
Added list of books to each user and fix BooksFirestoreRepository

Chockencore:
This PR adds a list of books to each user in the form of a list of UUID of the books. It replaced the temporary list of books for the map by a temporary list of user, but now it takes the books from the database according to the list of books of the users.  The list of user will be replaced when a similar file to BooksFirestoreRepository but for the users will be created.

In the middle of the development, it became apparent that the function documentToBooks from the BooksFirestoreRepository had some problem. The names to retrieve the information are fixed and the UUID is also taken from the database. Also some return null in case of certain parameters were null in the database were removed as in the databook, these parameters can be null.

Some tests needed to be fixed to continue working, like the end to end test and the login tests. Nothing was changed in the test, it was only the setup or the setcontent that was modified.


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -72,22 +74,23 @@ var SemanticsPropertyReceiver.cameraPosition by CameraPositionKey
...
 @Composable
 fun MapScreen(
-    listUser: List<TempUser>,
-    selectedUser: TempUser? = null,
+    listUser: List<DataUser>,
     navigationActions: NavigationActions,
     bookFilter: BookFilter,
-    geolocation: IGeolocation = DefaultGeolocation(),
+    booksRepository: BooksRepository,
+    selectedUser: Int = -1,
```
matiisalv, line 92: add a constant value

app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -104,17 +107,36 @@ fun MapScreen(
...
+        OnSucess = { books ->
+          listUser.forEach { user ->
+            userBooksList.add(
+                UserBooksWithLocation(
+                    user.longitude,
+                    user.latitude,
+                    books.filter { book -> book.uuid in user.bookList }))
+          }
+        },
+        onFailure = {})
```
matiisalv, line 124: could provide more information if there is an error with example an Log

matiisalv (approved):
Brilliant work! The code looks nearly perfect, with just a few minor fixes you can feel free to address. Overall, it’s very well done and looks excellent to me

app/src/main/java/com/android/bookswap/ui/map/Map.kt
Chockencore: I will do it in another PR as I am currently working on a branch that remove the logic for getting the list of books from the UI and put it in a model file.

PR #145, opened 2024-10-31
Feat/linkchatwithlistchat/matiisalv

matiisalv:
Add link between ListChatScreen and ChatScreen
Add viewModel to manage the connection between them


oliveja22 (approved):
I see no issues, it is concise and correct, it effectively does what it's supposed to, from the filtering in the listener to the new navigateTo which includes more parameters. It is simply well done I believe, nicely done.

PR #162, opened 2024-11-06
Refactor and enhance implementation of custom components

Chocolat2:
Refactor the implementation of existing custom components to ensure proper theming following Figma's design.
Overhaul the implementations to add or extend Material3 theming support and ensure Material3 design tokens are useable to take into account changes in theming without having to modify each component one by one.

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/components/FieldComponent.kt
+++ b/app/src/main/java/com/android/bookswap/ui/components/FieldComponent.kt
@@ -1,15 +1,53 @@
 package com.android.bookswap.ui.components
 
+import android.content.res.Configuration
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.interaction.InteractionSource
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.interaction.collectIsFocusedAsState
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.defaultMinSize
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.BasicTextField
+import androidx.compose.foundation.text.KeyboardActions
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material.ExperimentalMaterialApi
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.LocalTextStyle
+import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedTextField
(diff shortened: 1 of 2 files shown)


Chockencore (changes requested):
The PR has some problems : 
1. The buttons on the ListChatScreen to go to the ChatScreen doesn't work anymore in this branch.
2. The button search on the AddISBNScreen doesn't look like the previous button, see images below.
3. The preview shouldn't be in comments because the preview should be shown without modifying the file and the ktfmt formatting of the code remove unused import. Each time someone wants to see the preview, he needs to import again everything that the preview needs which is bothersome. If the problem is that the line coverage is reduced because of this, then leave it like that and crate a task to move the preview to a folder which isn't taken into account in the jacoco report.

Altered version:
Screenshot from 2024-11-12 15-38-29

normal version:
normalISBN

Chockencore:
The point 1 of my request changes isn't linked to this PR, this comes from another PR that was merged previously.

Chocolat2:
> The PR has some problems :
> 
> 1. The buttons on the ListChatScreen to go to the ChatScreen doesn't work anymore in this branch.
This most likely doesn't work on main either since the files changes are not related

> 3. The preview shouldn't be in comments because the preview should be shown without modifying the file and the ktfmt formatting of the code remove unused import. Each time someone wants to see the preview, he needs to import again everything that the preview needs which is bothersome. If the problem is that the line coverage is reduced because of this, then leave it like that and crate a task to move the preview to a folder which isn't taken into account in the jacoco report.
Previews are meant to be in the same file as the composable, this is an Android Studio feature to help in UI design. Moving them in another folder would defeat the purpose of having previews as a whole.

> 4. The values shouldn't be hard coded, they should be defined in constants
done

> 5. With the changes of the buttons component, the colors for the text of the ISBN text field is now black, the border of the text field changes colors and also the text field name "ISBN*" changes colors. If some of these changes are voluntary, then the figma should be modified to reflect these changes.
The label text color is managed automatically by the material theme. This color is dependant on the container color. Will update Figma

Chocolat2:
Figma updated to reflect the changes in design needed for accessibility purposes

Chockencore (changes requested):
Most of the requested changes were made but there are still some problems.
1. There are still hard coded values in the files that this PR modified, without counting the AddISBNScreen as the hard coded values were already there before this PR.
2. There are no more background color in the AddISBNScreen, meaning that only the top bar has the right color and the rest in on a white background.
3. The search button isn't the same as in the figma, it is the wrong color and the width isn't the same. The width is more of a detail.

app/src/main/java/com/android/bookswap/ui/components/FieldComponent.kt
```diff
@@ -18,26 +53,285 @@ fun FieldComponent(
...
+    colors: TextFieldColors,
+    focusedBorderThickness: Dp,
+    unfocusedBorderThickness: Dp
+): State<BorderStroke> {
+  val focused by interactionSource.collectIsFocusedAsState()
+  val indicatorColor = indicatorColor(enabled, isError, interactionSource, colors)
+  val targetThickness = if (focused) focusedBorderThickness else unfocusedBorderThickness
+  val animatedThickness =
+      if (enabled) {
+        animateDpAsState(targetThickness, tween(durationMillis = 150))
```
Chockencore, line 193: Here is a example of one of the hard coded value remaining

Chockencore (approved):
The PR is now good. The problem with the colors of certain elements will be resolved in a future task.

PR #169, opened 2024-11-07
UI Chat Picture Implementation for UI

oliveja22:
## Picture UI Implementation for Chat
This pull request consists mainly of the possibility of displaying an image in chat, it uses the previous composable `MessageItem` and adds a conditional to see if the ID is "101", which is the string currently used to check for it, this will change. It also includes additional features that are related:
- Pop up with larger photo when clicking on it, allowing a clearer vision of a picture which is usually quite small
- Zoom option within the pop up, allowing for zooming in, out and moving the picture as needed.

While this looks good, there are additional changes related to further tasks which are described below:
- Change MessageData to use an enum type instead of ID, which is a momentary test #168
- Implement a usage of URL instead of a placeholder image #167


Chockencore (changes requested):
The PR is quite good but there is some problem.
1. The values shouldn't be hard coded, it should be in constants.
2. When we click on the edit button for the image, it gives us a url, which is strange. There is two possibility to fix that, the first is to just remove the edit button, so we can only delete images. The second is to add later on a functionnality to change the desired image but as we still didn't make the base for adding a image, it will be just a TODO. The edit button would do nothing for now. Nonetheless, the code should differentiate between a image and a text for a longPress.
3.  When clicking on the image, the box that contains the zoomed image is a square, meaning that we can interact with the image even when clicking next to it. It should be the same dimension as the image.
4. When dragging the zoomed image, it is dragged faster than what is really dragged. It travels a greater distance than it should be.
5. I would say that the dragged feature souldn't be active when the image is smaller than the screen, and the image should be stuck to the border of the screen. Right now, it is possible to drag the image outside of the screen.

app/build.gradle.kts
```diff
@@ -136,6 +136,7 @@ dependencies {
     implementation("com.google.firebase:firebase-auth")
     implementation("com.google.firebase:firebase-analytics")
     implementation(libs.google.play.services.location)
+    implementation("io.coil-kt:coil-compose:2.1.0")
```
EdenKahane, line 139: Avoid adding dependency version in this file directly

EdenKahane (approved):
This looks very professional ! 
The additional tests are well made and seems to test everything relevant.
The new image UI is really clean too.

Make sure to remember to delete the placeholders image when linking this to the file uploader.

oliveja22:
In accordance to [Chockencore] (https://github.com/Chockencore)'s review, I have a few things to add, panning has been removed, as it was never intended to take place, only zoom, fixing 4 and 5 in their list. Additionally 2 belongs to #168 , apologies for the lack of detail in that regard. The other two should be fixed now.

PR #172, opened 2024-11-07
Sort the users by the distance from the user and moved logic from the mapScreen

Chockencore:
This PR moved the logic for retrieving the books and filtering them, from the map screen to a new model file named BookManager. As a few of the argument for the map screen is only used by the BookManager, these arguments were replaced by the book manager.
This PR  also add in the BookManager a sorting system that sort the books by the distance between their owner and the user's geolocation.

This PR doesn't have currently the test for the BookManager, so it shouldn't be merged.

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
@@ -19,71 +19,88 @@ import com.android.bookswap.data.DataBook
 import com.android.bookswap.data.DataUser
 import com.android.bookswap.data.source.network.BooksFirestoreRepository
 import com.android.bookswap.model.map.BookFilter
+import com.android.bookswap.model.map.BookManager
 import com.android.bookswap.model.map.DefaultGeolocation
 import com.android.bookswap.ui.navigation.NavigationActions
 import com.google.maps.android.compose.CameraPositionState
 import io.mockk.every
+import io.mockk.just
 import io.mockk.mockk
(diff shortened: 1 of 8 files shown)


TheoSchlaeppi (changes requested):
Thank you for this PR; it’s necessary and well-written. However, I believe there are several possible improvements.

Firstly, from what I understand, the `while(true)` loop in `BookManager` runs continuously, which could lead to optimization issues. Using a flag to track when the user is on the map screen could be a good approach to address this by refreshing data only when necessary. Additionally, the error handling doesn’t allow for an immediate retry, requiring a 5-second wait before the next attempt. Improving this retry mechanism would enhance the user experience by reducing the delay when books don’t load right away.

Regarding the architecture, I can say that BookManager is indeed acting as a ViewModel, managing UI state and orchestrating data flows. It correctly follows MVVM principles in this regard. It handles data transformation (filtering books, calculating distances) and exposes state via StateFlow for the UI to observe.

On a positive note, there are many solid aspects in this code. The use of StateFlow to handle and share data updates is excellent and fits well within a reactive framework. The structure is also clean and organized, making it easy to follow and extend. With a few adjustments, this code could be even more optimized and aligned with best practices. Great work overall! (I didn't mention it but test are good )

TheoSchlaeppi (approved):
Thank you for considering my suggestions, including implementing the retry mechanism and stopping the loop when the user is no longer on the screen. I believe the code is solid, and I’m happy to approve it. Thanks for your commitment to the project!

PR #176, opened 2024-11-08
Refractor the whole app so all data type use UUID

EdenKahane:
Before this, depending on the datatype and the data source, we were sometimes using strings, sometimes UUID, generated by firestore or sometimes randomly, sometimes with the name "id" sometimes with the name "uid", etc...
Now everything use UUID and follow the same code structure.

Due to some obscure test logic it took way longer than expected. Especially, I was twice forced to rewrite all the tests of a file because the tests were 500 lines long when ~80 was enough for the same coverage or because the wrong thing was tested.

Refactoring was not enough on most of the change, since I also needed to locate and change the logic of data parsing etc... 

In the end, I didn't fix everything that could be fixed, just the things that were too painful to left as it is.
I also created a few tasks for future improvement.


EdenKahane:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=BookswapEPFL_Bookswap&pullRequest=176) **Quality Gate failed**
> 
> Failed conditions  [79.22% Line Coverage on New Code](https://sonarcloud.io/dashboard?id=BookswapEPFL_Bookswap&pullRequest=176) (required ≥ 80%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=BookswapEPFL_Bookswap&pullRequest=176)

This doesn't make sens for this PR since the problem is each time on just line changes. Or like 1-2 lines added.

PR #180, opened 2024-11-08
Feature/matiisalv/photoview model

matiisalv:
This PR contains a ViewModel that manage the interaction with PhotoRepository.
And change the way of  how the photo are managed in the book
Some Test are needed , but will be incorporate when the view model we will part of the others module


app/build.gradle.kts
```diff
@@ -169,6 +169,7 @@ dependencies {
     implementation(libs.firebase.auth.ktx)
     implementation(libs.androidx.navigation.compose)
     implementation(libs.firebase.messaging.ktx)
+    implementation(libs.androidx.uiautomator)
```
EdenKahane, line 172: What is this ? Why are you adding it ?

EdenKahane (changes requested):
No idea why you use `implementation(libs.androidx.uiautomator)`
Apart from that there is a lot of very useful things here.
It's not obligatory but it's better next time to use mockk instead of mockito.

matiisalv:
After we discuss  during the meeting this part we will not needed so it will not be merge in the code

PR #187, opened 2024-11-08
Feature/leni cadorin epfl/matiisalv/photo management

matiisalv:
This PR is introduce the interface of PhotoRepository.
This Interface helps to work in different Branch and help future test

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/DataPhoto.kt
+++ b/app/src/main/java/com/android/bookswap/data/DataPhoto.kt
@@ -8,4 +8,5 @@ data class DataPhoto(
     val timestamp: Long,
     val userId: String,
     val bookId: String,
-    val uri: Uri? = null,)
+    val uri: Uri? = null,
+)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/bookswap/data/DataPhoto.kt
```diff
@@ -0,0 +1,12 @@
+package com.android.bookswap.data
+
+import android.net.Uri
+
+data class DataPhoto(
+    val id: String,
+    val url: String,
+    val timestamp: Long,
+    val userId: String,
+    val bookId: String,
```
oliveja22, line 10: We need to discuss whether the bookId is necessary, but it should be fine for the moment, as a picture doesn't necessarily have to be related to an image, but it could always be null

oliveja22 (approved):
Everything is quite good here, a small comment but not necessary to change anything, as a result, I can approve it.

PR #192, opened 2024-11-11
Link Profile Screen to app navigation

Chocolat2:
Link the User's Profile Screen to the App Navigation.

- MainActivity.kt:
  - Add the Profile Screen to the Navigation Host's navigation tree
  - Fix various parameter types
  - Fix some comments
- NavigationActions.kt:
  - Add the Profile Screen path strings to the Route and Screen objects
  - Add the Profile Screen to the ToplevelDestinations object
- UserProfileScreenTest.kt:
  - Remove some unused imports and annotations


Chockencore (changes requested):
This PR is great overall. The only thing to be added to this PR would be to add to the profile Icon the navigation to the profile page as it is the only way the profile screen should be accessed and it would be a bit redundant to create a PR just for this.

Chocolat2:
> This PR is great overall. The only thing to be added to this PR would be to add to the profile Icon the navigation to the profile page as it is the only way the profile screen should be accessed and it would be a bit redundant to create a PR just for this.

Check #193

Chockencore (approved):
Nevermind, the requested change is treated in the mentioned PR by Chocolat2.

PR #193, opened 2024-11-11
Implement custom TopAppBar component

Chocolat2:
Implement a custom TopAppBar component to use proper style and theme.
- Create the custom TopAppBar component
- Replace any TopAppBar in the current code with the custom component
- Ensure proper functionality after the refactor and fix any problems


oliveja22 (approved):
I can see no major issues, it seems fine and although so many files have been modified, they seem to be 2 to 3 lines at most. It seems correct, and I cannot see much fault with it.

PR #194, opened 2024-11-11
 Prevent unnecessary screen reloads when navigating to the current route

LeniCadorinEpfl:
This PR addresses an issue in the app’s navigation where pressing a button to navigate to the current screen (e.g., tapping the map button while already on the map screen) would cause the screen to reload. This unintended behavior led to a brief black screen as the screen reloaded unnecessarily.

**Changes:**
Added a check in `NavigationActions` to ensure navigation occurs only if the target route is different from the current route.

**Tests:**
New unit tests have been added in `NavigationCurrentLocation.kt` to validate the changes


TheoSchlaeppi (approved):
Thank you for this PR, which clearly enhances the UX in a very simple and elegant way. You've also added tests for this functionality, which is great. Everything looks correct to me, and I fully approve these additions.

PR #198, opened 2024-11-12
DataMessage refactoring to include message type

oliveja22:
### Refactoring DataMessage to include message type
Create an enum with types `TEXT` and `IMAGE`, then update all necessary files. The new DataMessage is as follows:
```kotlin
data class DataMessage(
    val messageType: MessageType = MessageType.TEXT,
    val uuid: UUID = UUID.randomUUID(),
    val text: String = "",
    val senderUUID: UUID,
    val receiverUUID: UUID,
    val timestamp: Long = 0L
)

enum class MessageType {
  TEXT,
  IMAGE
}
```

Then the chat screen and the firestore source files are updated with their respective tests. Another change is the usage of the type to check what message bubble will be used in the chat screen instead of a temporary id.


TheoSchlaeppi (approved):
Thank you for this PR! For me, it's looking good and I let you merge it.

Issue #175, opened 2024-11-08
Refactor things related to BooksRepository and its tests.


EdenKahane:
- [x]  BooksFirestoreRepository should be a source
- [x] Document for functions should use /** */ 
- [x] Avoid nested tests in BooksFirestoreRepositoryTest
- [x] No longer use onSuccess and onFailure but just "callback"
- [x] Pass from uuid.mostSignificantBits to just a normal uuid storage

PR #199, opened 2024-11-12
fix!: Fix BooksRepository and source

EdenKahane:
See #175


EdenKahane:
Yet again, not adding anything new, just doing some refactoring. So the actual line coverage doesn't change.

PR #201, opened 2024-11-13
Fix navigation to Chat

oliveja22:
## Fix navigation to chat
This pull request seeks to fix the issue arisen with the new implementation of UUIDs everywhere that impeded accessing individual chats due to an `IllegalArgumentException` when changing from usernames to UUID. This has been fixed by implementing the use of `DataUser` in `ListChat` and `ChatScreen`.
The new `ChatScreen` looks as follows (to give an example):
```kotlin
fun ChatScreen(
    messageRepository: MessageRepository,
    currentUser: DataUser,
    otherUser: DataUser,
    navController: NavigationActions
) {
   //content
}
````
As it can be seen, they use DataUser now. Additionally, the navigation has been updated both in the main activity and the `NavigationActions`. Furthermore, a current user has been added to the main activity, which for the moment I have filled with my own information, such that it can be used, but it should be updated upon a login to be the right one.


matiisalv (approved):
Really good job, thanks you to take the time to do this, it s very beneficial  to all team, forme all look perfect

PR #204, opened 2024-11-14
Feature/matiisalv/documationadded

matiisalv:
This PR add documentation to all  the program in DocString format


app/build.gradle.kts
```diff
@@ -169,6 +169,7 @@ dependencies {
     implementation(libs.firebase.auth.ktx)
     implementation(libs.androidx.navigation.compose)
     implementation(libs.firebase.messaging.ktx)
+    implementation(libs.androidx.uiautomator)
```
EdenKahane, line 172: What is this ? Why is it added here ?

EdenKahane (changes requested):
The document is great and much needed  !
I do not understand why you add `implementation(libs.androidx.uiautomator)`, I think it's just a temporary test and should be removed.

PR #213, opened 2024-11-14
Fix the documentToBooks in the BooksFirestoreSource

Chockencore:
This PR fixes the problem of the documentToBooks using the wrong method to retrieve the uuid of the book. It uses the old way of handling the uuid.


matiisalv (approved):
Look's good for me

PR #217, opened 2024-11-14
Test implement an End to End test for the add books feature

TheoSchlaeppi:
This PR adds end-to-end tests for the book addition feature, including verification of the ISBN-based search functionality and manual entry screen. When the option to add a book via photo becomes available, these tests will need to be extended to cover that functionality as well. (also added one testTag for the genre of the book)

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/endtoend/AddBooksEndToEnd.kt
+++ b/app/src/androidTest/java/com/android/bookswap/endtoend/AddBooksEndToEnd.kt
@@ -1,0 +1,125 @@
+package com.android.bookswap.endtoend
+
+import androidx.compose.ui.test.hasText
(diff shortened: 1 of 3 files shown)


oliveja22 (approved):
While I will approve it, it would be nice if you could insert comments in your test explaining what each part of the test does. Other than that it is very clean and I see no other issues with it.

Chocolat2 (approved):
Overall code looks good although improvements can be made.

Line comments are somewhat minimal but still enough to understand the code.
Some improvements that can be made regarding comments in general are:
- Add javadoc style comment to each function with a description of the parameters and return valies, if any, and a summary of the internal computation and useage. (Check PR #221 for examples)
- Normalize where line comments are written, either inline or directly above the line

PR #244, opened 2024-11-19
Fix the book manager view model

Chockencore:
This PR fixes the problem of certain CustomInfoWindows that didn't appear. It was because the book manager's list of user also contained the user without books, so the selectedUser value pointed to the wrong user. They were removed of this list as they should be filtered out for the map screen. 
This PR also removed the problem of the update logic of book manager. When stopping the updates, the book manager killed the coroutine so it couldn't be restarted. Therefore, once the user goes away of the map screen to go to another screen like the filter map screen, the book manager wouldn't start updating again after going back to the map screen. It was fixed by canceling the jobs of the coroutine and not the coroutine.


TheoSchlaeppi (changes requested):
Thank you for this elegant and necessary fix. Your code is well-written, but there’s a small typo in the variable name `fetchBooksFromRepository`. It should end with `Job` to maintain consistency with the other jobs. Your code updates the tests, so in my opinion, this code is a welcome addition. 

Please fix this small detail and contact me as soon as the modification is made, so I can approve your excellent work.

TheoSchlaeppi (approved):
Looks good for me !

PR #245, opened 2024-11-19
Feature/matiisalv/mapscreen off line

matiisalv:
This PR, have for Purpose to enable the MapScreen, when the user is offline, and for introduce a function named  isNetworkAvailable , that test if the user can be connected to internet

### DIFF
--- a/app/src/main/java/com/android/bookswap/model/NetoworkOnline.kt
+++ b/app/src/main/java/com/android/bookswap/model/NetoworkOnline.kt
@@ -15,5 +15,6 @@ fun isNetworkAvailable(context: Context): Boolean {
       context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
   val network = connectivityManager.activeNetwork ?: return false
   val capabilities = connectivityManager.getNetworkCapabilities(network) ?: return false
-  return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+  return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) &&
+      capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
 }
--- a/app/src/main/java/com/android/bookswap/model/NetoworkOnline.kt
+++ b/app/src/main/java/com/android/bookswap/model/NetoworkOnline.kt
@@ -1,0 +1,13 @@
+package com.android.bookswap.model
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+
+fun isNetworkAvailable(context: Context): Boolean {
+  val connectivityManager =
+      context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -96,116 +100,121 @@ fun MapScreen(
...
+    isOnline = isNetworkAvailable(context)
   }
-  // Stop location and books updates when the screen is disposed
-  DisposableEffect(Unit) {
-    onDispose {
-      geolocation.stopLocationUpdates()
-      bookManagerViewModel.stopUpdatingBooks()
+  if (!isOnline) {
+    Text("No internet connection")
+  } else {
```
Chocolat2, line 107: This condition will compose a blank screen with a simple Text when offline, without the navigation and top app bar, and prevents the user from using the app in offline mode entirely.
The only useage of internet on this screen is when calling `bookManagerViewModel.startUpdatingBooks()`. Thus when offline, we can simply skip this call and display a Toast message such as "No internet connexion" instead.

app/src/main/java/com/android/bookswap/model/NetoworkOnline.kt
Chocolat2, line 1: Code looks good but lacks comments.
The file name and package may not be the best for this utility function, `com.android.bookswap.utils` seems to be a more appropriate package and renaming the file to `NetworkUtils.kt` would better describe the file content and useage

app/src/test/java/com/android/bookswap/AvaibleNetworkTest.kt
Chocolat2, line 1: This test suite seems compleat
More comments could be added to explain the mocks and tests in more details

app/src/main/java/com/android/bookswap/ui/map/Map.kt
Chocolat2, line 1: The current implementation is breking the app entirely, needs fixing

Chocolat2 (changes requested):
Changes on the Map.kt file break the app!
To fix ASAP

app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -96,125 +97,125 @@ fun MapScreen(
...
-        CameraPosition.fromLatLngZoom(LatLng(latitude.value, longitude.value), INIT_ZOOM)
+    if (isOnline) {
+      bookManagerViewModel.startUpdatingBooks()
+      geolocation.startLocationUpdates()
+      cameraPositionState.position =
+          CameraPosition.fromLatLngZoom(LatLng(latitude.value, longitude.value), INIT_ZOOM)
+    } else {
+      Toast.makeText(context, "Please connect to Internet to actualise", Toast.LENGTH_SHORT).show()
+    }
     isOnline = isNetworkAvailable(context)
```
Chocolat2, line 108:
```suggestion
```
This line is a functional duplicate of §92 and serves no purpose. When the screen is first composed, all the remembered values are computed and only after that are the components and LaunchedEffects processed, so the value of `isOnline` is already computed at §92

Chocolat2 (changes requested):

PR #248, opened 2024-11-21
Implement navigation to and from `EditBookScreen'

LeniCadorinEpfl:
This PR implement the navigation to and from the `EditBookScreen`. It also introduces  improvements to manage user-specific book data and provides UUID-based book retrieval and modification. The following changes are included:

### 1. Navigation Enhancements:
- Introduced navigation logic between `BookProfileScreen` and `EditBookScreen`.
- Updated `MainActivity` to handle bookId and user-specific context for accurate book operations.

### 2. User-Specific Book Management:
- Refactored the book addition and modification processes to include a userId parameter.
- Updated the PhotoData model to incorporate userId, ensuring proper association between users and their books.

### 3. Repository Improvements:
- Added a new getBook(uuid: UUID) method in `BooksFirestoreRepository` to retrieve books using UUIDs.
- Enhanced existing methods (addBook, updateBook) to support UUID and user-specific contexts in Firestore.
### 4. Component Modifications:
- `BookProfileScreen`:
  - Modified function signature to accept bookId and booksRepository instead of DataBook (This change is necessary to ensure the correct book data is retrieved from the database when modifying book values.)
  - Ensures book data is fetched dynamically based on UUID and user context.
- `EditBookScreen`:
  - Updated to support direct book editing using UUID and userId.
- `AddToBookScreen` & `AddISBNScreen`:
  - Integrated userId parameter to manage user-specific book data during creation.

### 5. Testing Improvements:
- Tests for:
  - Navigation from `BookProfileScreen` to `EditBookScreen`.
  - Update each book addition test to support userid .


app/src/main/java/com/android/bookswap/ui/navigation/NavigationActions.kt
```diff
@@ -89,14 +91,32 @@ open class NavigationActions(
...
+      val route =
+          when (screen) {
+            Screen.EDIT_BOOK -> "$screen/$UUID"
+            else -> screen
+          }
       navController.navigate(route)
     }
   }
 
+  /*
```
oliveja22, line 111: is it necessary to keep this comment here?

app/src/test/java/com/android/bookswap/data/source/network/BooksFirestoreRepositoryTest.kt
```diff
@@ -0,0 +1 @@
+
```
oliveja22, line 1: What happened here? Why is there an empty file like this?

oliveja22 (changes requested):
There pull request is mostly very good, most of the changes are of course adapting a number of files that have small changes as a result. However, there are three things to note, two of which would have nonetheless resulted in me approving; those are the unnecessary comment as far as I can see and the multiple usage of a map which can be made into a function. All of these are shown as a message in the files. There is a problem however which I feel needs change. There is an empty file that has seemingly been modified, i.e it was there before, it is not a file which has just been added. This means to me that there was something there at some point, even if it doesn't show it. If it is completely unnecessary please remove it, it is app/src/test/java/com/android/bookswap/data/source/network/BooksFirestoreRepositoryTest.kt. If it is not an issue and not important, please just delete it.

LeniCadorinEpfl:
Thank you for the detailed review and feedback! Here's how I addressed the points you mentioned:

**1. Unnecessary Comment in `NavigationActions`:** I have removed the commented section as it was indeed unnecessary.

**2. Usage of bookMap:** I decided not to create a separate function for bookMap as it is only used once in the code. As I noticed it was not used in `updateBook`, so I removed it.

**3. `BooksFirestoreRepositoryTest.kt` File:** I have deleted this file as it was a result of a merge conflict. The file name had changed to `BooksFirestoreSourceTest.kt` in another branch, but during the merge, the old file was not removed. This has now been resolved.

Thank you again for pointing these out! Let me know if there’s anything else I should revisit.

oliveja22 (approved):
All changes seem to match what I asked for, so it looks great and I shall approve it.

PR #250, opened 2024-11-21
Offline Messages Implementation

oliveja22:
## Implement offline message storage
This pull request seeks to implement a method for storing a certain amount of texts in a way that they can be accessed online. For this, this pull request includes the following new files:

- `Messages.txt` x2 (for storing the texts, one for tests and the other for the actual usage)
- `OfflineMessageStorage.kt`
- `OfflineMessageStorageTest.kt`

These files seek to implement said method. The way it works is by adding in the `OnLaunchEffect` of `ChatScreen` a call to a function which takes the messages that have already been taken from firebase and then calling the function to add them to the file. Furthermore, it calls the function to request them in the callback given to the function to retrieve from firebase upon a failure. This way when it fails because of whatever reason, it can call to the offline file. Additionally, and unintended, so I do not know if this is only for my android phone, but it seems to store in a cache of some sort the images taken from URLs, so sometimes it can even load the images while offline.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -155,6 +155,7 @@ dependencies {
     implementation ("com.google.firebase:firebase-storage:20.1.0")
     implementation(libs.google.play.services.location)
     implementation("io.coil-kt:coil-compose:2.1.0")
+    implementation("com.google.code.gson:gson:2.8.9")
     implementation(libs.androidx.core.ktx)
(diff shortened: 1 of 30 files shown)


LeniCadorinEpfl:
The new encryption and decryption scheme works perfectly, it is good to go for merging, in my opinion!

LeniCadorinEpfl (approved):

PR #252, opened 2024-11-21
Feature/matiisalv/add screen off line

matiisalv:
This PR add Offline Mode to the AddToBookScreen

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/map/Map.kt
+++ b/app/src/main/java/com/android/bookswap/ui/map/Map.kt
@@ -34,6 +34,7 @@ import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.SemanticsPropertyReceiver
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -86,16 +89,23 @@ fun MapScreen(
...
-    geolocation.startLocationUpdates()
-    cameraPositionState.position =
-        CameraPosition.fromLatLngZoom(LatLng(latitude.value, longitude.value), INIT_ZOOM)
+    if (isOnline) {
+      bookManagerViewModel.startUpdatingBooks()
+      geolocation.startLocationUpdates()
+      cameraPositionState.position =
+          CameraPosition.fromLatLngZoom(LatLng(latitude.value, longitude.value), INIT_ZOOM)
+    } else {
+      Toast.makeText(context, "Please connect to Internet to actualise", Toast.LENGTH_SHORT).show()
```
oliveja22, line 106: the correct word you are looking for in this toast is update, not actualise

oliveja22 (approved):
I find no particular issues with this code, there is not much and it is simple, but effective. I would not have thought of making a whole function which checks whether there is connectivity to the internet, so I think it is quite good.

app/src/main/java/com/android/bookswap/ui/map/Map.kt
Chocolat2: I would've shown either "There was a network error" or "Network has no internet access" for the text since these are strings already used in the Android API.
Both are found in Android API > res > values > strings.xml, with keys `httpError` and `other_networks_no_internet` and in the translated values using the same keys.

Using strings already used in Android lets us copy-paste translations rather than do the translation ourselves and also ensures proper string length for accessibility.

If needed, a dismissable snackbar can be used instead of a toast message to display additional info concerning the error, using a modal dialogue shown when the snackbar is clicked.

Chocolat2 (changes requested):
This branch is based off of an old commit from #245 [76b2dd7d](https://github.com/BookswapEPFL/Bookswap/pull/245/commits/76b2dd7d0729593b2cafeaca7a50fb76be7224fe)
Requested changes and fixes made after 76b2dd7d **MUST** be merged into this branch before merging anything into main.

PR #270, opened 2024-11-25
Add a button to the Map screen to recenter the camera on the current user's location

Chocolat2:
Enable Google Map's "My Location" button


EdenKahane:
This is very useful. 
Nice work using the ui setting insteading of recoding the thing manually.
And the GrantPermissionRule tricks will be very useful for later !

EdenKahane (approved):
This is very useful.
Nice work using the ui setting insteading of recoding the thing manually.
And the GrantPermissionRule tricks will be very useful for later !

PR #272, opened 2024-11-26
Add information for the DataBook for knowing when it's exchange and archived

matiisalv:
This PR to element at the DataBook, to know  when the book is exchange or archived


app/src/androidTest/java/com/android/bookswap/ui/books/edit/EditBookTest.kt
```diff
@@ -148,16 +149,31 @@ class EditBookScreenTest {
         .onNodeWithTag(C.Tag.EditBook.language)
         .assertTextContains(sampleBook.language.toString())
   }
-
+  /*
   @Test
```
oliveja22, line 153: Why is this test commented? If it doesn't work, one cannot simply commented as such, if it is no longer needed as other tests have covered what this test once covered, then please let it be removed rather than staying as a comment.

app/src/main/java/com/android/bookswap/data/source/network/BookFirestoreSource.kt
```diff
@@ -151,10 +156,21 @@ class BooksFirestoreSource(private val db: FirebaseFirestore) : BooksRepository
     Log.d("BooksFirestoreRepository", "Attempting to add book: ${dataBook.title}")
 
     // Attempt to add book to Firestore
-    val bookDocument = bookToDocument(dataBook)
+    val bookMap =
```
oliveja22, line 159: The usage of the bookToDocument function is better and should probably be maintained here, as it can be used further in other functions. Given that a DataBook is passed as a parameter and used for the map, and given that the DataBook now consists of all previous variables with the addition of the two new ones, I cannot understand why you did not maintain the usage of bookToDocument; this is especially true given that you modified bookToDocument in order to account for these new variables. If this is due to a misunderstanding on my part, please excuse it and hopefully given an explanation I may understand the absence of those two variables being mapped.

app/src/test/java/com/android/bookswap/data/source/network/BooksFirestoreSourceTest.kt
```diff
@@ -111,25 +114,86 @@ class BooksFirestoreSourceTest {
...
 
     // Verify Firestore update operation
-    verify { mockDocumentReference.set(bookSource.bookToDocument(testBook)) }
+    verify { mockDocumentReference.set(bookDocument) }
   }
 
   @Test
   fun `documentToBooks is valid`() {
-    val bookSource = BooksFirestoreSource(mockFirestore)
+    // Arrange
```
oliveja22, line 145: Here you should use the testBook that you created in the @Before, as it is almost identical and would certainly serve the same purpose

oliveja22, line 145: Here you should use the testBook that you created in the @Before, as it is almost identical and would certainly serve the same purpose

oliveja22 (changes requested):
Overall the pull request is quite good, it implements the two new variables in the DataBook thoroughly and mostly correctly, there are a few changes that have to be made however. The first is that there is a commented test that should either be reactivated and fixed if it fails or removed. The second and last is that in BooksFirestoreSource you use a slightly different map which does not take into account the two new variables instead of using the map you created. This seems wrong to me, but like I said in the comment, if this is a wrong assessment, please just explain it to me and I will approve.

app/src/main/java/com/android/bookswap/data/DataBook.kt
```diff
@@ -3,16 +3,20 @@ package com.android.bookswap.data
 import java.util.UUID
 
 /**
- * Represent a book with various properties
+ * Represents a book with various properties.
  *
```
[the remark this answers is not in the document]
matiisalv: thanks

oliveja22 (approved):
Thanks for taking into account all I asked for. I will hence here approve it.

PR #275, opened 2024-11-28
Add common AppConfig available in all the Components

EdenKahane:
As indicated by [this](https://developer.android.com/develop/ui/compose/compositionlocal?hl=fr).
Android Compose provides a simple way for developer to access common data from any component.

AppConfig can store all the data that should be stay the same across the whole app.

data can be accessed using `val config = LocalAppConfig.current` in a composable environment.

This PR does not pass SonarCloud, this is normal see [this](https://github.com/BookswapEPFL/Bookswap/pull/275#issuecomment-2505081421)


EdenKahane:
Yet again, for SonarCloud, only 5 lines are not tested.
These are just the component added by Jetpack compose, so not really testable.

PR #276, opened 2024-11-28
Implement new contact from new chat functionality

oliveja22:
## Contact gets added when new chat is created
This PR is quite simple, to the file `UserFirestoreSource.kt` it adds two methods, one to add a contact to the contact list of a given UUID and another one to remove the contact. It also adds the necessary tests. Additionally, in `ChatScreen.kt`, in the effect on launch, it tests if the list of messages is longer than 0, and if so, it checks if the contacts already exist within each others  `DataUser`. If they don't exist, they get added.

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/repository/UserRepository.kt
+++ b/app/src/main/java/com/android/bookswap/data/repository/UserRepository.kt
@@ -61,4 +61,22 @@ interface UsersRepository {
    *   Result.failure(exception) if error
    */
   fun deleteUser(uuid: UUID, callback: (Result<Unit>) -> Unit)
+
+  /**
+   * Function to add a contact to the user's contact list in the repository.
+   *
+   * @param userUUID UUID of the user whose contact list is being updated.
+   * @param contactUUID UUID of the contact to add.
+   * @param callback Callback for success or failure.
+   */
+  fun addContact(userUUID: UUID, contactUUID: String, callback: (Result<Unit>) -> Unit)
+
+  /**
+   * Function to remove a contact from the user's contact list in the repository.
+   *
+   * @param userUUID UUID of the user whose contact list is being updated.
+   * @param contactUUID UUID of the contact to remove.
+   * @param callback Callback for success or failure.
+   */
+  fun removeContact(userUUID: UUID, contactUUID: String, callback: (Result<Unit>) -> Unit)
 }
--- a/app/src/main/java/com/android/bookswap/data/source/network/UserFirestoreSource.kt
+++ b/app/src/main/java/com/android/bookswap/data/source/network/UserFirestoreSource.kt
@@ -179,4 +179,62 @@ class UserFirestoreSource(private val db: FirebaseFirestore) : UsersRepository {
       }
     }
   }
+
+  /**
+   * Adds a contact to the user's contact list in Firestore.
+   *
+   * @param userUUID UUID of the user whose contact list is being updated.
+   * @param contactUUID UUID of the contact to add.
+   * @param callback Callback for success or failure.
+   */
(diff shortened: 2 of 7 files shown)


oliveja22:
Next Week I will do an end to end that tests the onLaunchEffect which is where the tests are lacking, hence why the low line coverage. Please ignore it for now.

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
```diff
@@ -173,6 +175,56 @@ fun ChatScreen(
...
+                      "ChatScreen",
+                      "Failed to add contact: ${contactResult.exceptionOrNull()?.message}")
+                }
+              }
+            }
+          } else {
+            Log.e(
+                "ChatScreen", "Failed to fetch current user: ${result.exceptionOrNull()?.message}")
+          }
+        }
```
Chocolat2, line 203: The user currently logged in the app is already stored in the UserViewModel so this call can most likely be simplified as `UserViewModel.getUser(force = true)`
On top of that, the current implementation doesn't comply with the MVVM template as this is calling data from the UI.

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
```diff
@@ -173,6 +175,56 @@ fun ChatScreen(
...
+                  Log.e(
+                      "ChatScreen",
+                      "Failed to add contact: ${contactResult.exceptionOrNull()?.message}")
+                }
+              }
+            }
+          } else {
+            Log.e("ChatScreen", "Failed to fetch other user: ${result.exceptionOrNull()?.message}")
+          }
+        }
```
Chocolat2, line 226: This implementation doesnt comply with the MVVM template as this is calling data from the UI and should be changed as soon as possible.

app/src/test/java/com/android/bookswap/data/source/network/UserFirestoreSourceTest.kt
Chocolat2, line 1: The added tests are good but the tested functionality should be moved to the viewModel instead of relying on database queries

app/src/main/java/com/android/bookswap/data/source/network/UserFirestoreSource.kt
Chocolat2, line 1: The added functionality should be moved to the viewModel instead of relying on database queries

app/src/main/java/com/android/bookswap/data/repository/UserRepository.kt
Chocolat2, line 1: The added functionality should be moved to the viewModel instead of relying on database queries

Chocolat2 (commented):
The new functionality is good but the implementation doesn't follow the MVVM template and the data processing should be moved out of the UI.

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
oliveja22: Next week, when I fix the listener, I will change the whole OnLaunchEffect, I can do this then

app/src/main/java/com/android/bookswap/ui/chat/ChatScreen.kt
oliveja22: same as last time

app/src/main/java/com/android/bookswap/data/repository/UserRepository.kt
oliveja22: When the listener is fixed next week, all onLaunchEffect functionality will be moved to a different file to comply with MVVM, I can then fix this next week

matiisalv (approved):
Fantastic work as usual, everything look great for me , maybe check if the coverage it's sufficient , but t should be ok

PR #277, opened 2024-11-28
Implement translation from geocoords latitude and longitude to an address

Chocolat2:
Implement a translation from geocoords latitude and longitude to an address so that an actual address is displayed on the user's profile rather than the latitude and longitude of the user.
This implementation uses the Places API from Google Maps

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -189,6 +189,7 @@ dependencies {
     implementation(libs.play.services.maps)
     implementation(libs.maps.compose)
     implementation(libs.maps.compose.utils)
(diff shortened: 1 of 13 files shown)


EdenKahane (changes requested):
Currently, this PR modify way too much things and mostly things out of scope of this PR.
This should be split in multiple PR and the PR for geocoords should only contain change directly related to this change.

Chocolat2:
This PR is being reworked in PR #300 and will not be further updated.
Changes unrelated to geolocalisation translation will be added in a separate PR in the future.

PR #278, opened 2024-11-28
Fix method of storing data into Firebase

Chocolat2:
Fix method of storing data into Firebase to avoid inconsistencies in data formatting and/or loss of information

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
+++ b/app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
@@ -63,23 +63,16 @@ class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageReposit
   /**
    * Sends a message to the Firestore database.
    *
-   * @param message The message to be sent.
+   * @param dataMessage The message to be sent.
    * @param callback Callback to be invoked with the result of the operation. The result is a
    *   success if the message is sent successfully, or an exception on failure.
    */
-  override fun sendMessage(message: DataMessage, callback: (Result<Unit>) -> Unit) {
-    val messageMap =
-        mapOf(
-            "uuid" to message.uuid.toString(),
-            "text" to message.text,
-            "senderUUID" to message.senderUUID.toString(),
-            "receiverUUID" to message.receiverUUID.toString(),
-            "timestamp" to message.timestamp,
-            "messageType" to message.messageType.name)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
```diff
@@ -229,64 +221,53 @@ class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageReposit
...
+          if (task.isSuccessful) {
+            val document = task.result
+            if (document != null && document.exists()) {
+              val existingMessage = documentToMessage(document).getOrNull()
+              if (existingMessage != null) {
+                val currentTime = System.currentTimeMillis()
+                if (currentTime - existingMessage.timestamp <= fifteenMinutesInMillis) {
+                  val messageMap = messageToDocument(dataMessage)
+                  db.collection(COLLECTION_PATH)
+                      .document(dataMessage.uuid.toString())
```
oliveja22, line 247: Here you have reverted to the previous way of looking in the firebase for a message, it will not work, it is probably an error which arose when merging with main. Please have a look at it.

app/src/test/java/com/android/bookswap/data/source/network/DataConverterTest.kt
```diff
@@ -0,0 +1,80 @@
...
+  fun parseRawLong() {
+    val raw_long = "1234567890123456789"
+    val big_long: Long = 1234567890123456789
+    val parsed_l = DataConverter.parse_raw_long(raw_long)
+    assertEquals("Value $raw_long was not parsed correctly!!", big_long, parsed_l)
+  }
+
+  @Test
+  fun parseRawLongList() {
+    val big_long: Long = 1234567890123456780
```
oliveja22, line 38: There are a number of reused variables here, you might want to establish them at the top

oliveja22 (changes requested):
There are a few issues, which is to be expected in a pull request of such gargantuan size. First I must thank everything you have here done, as it is quite good and helpful, fixing many issues. However, I find myself having to mention two issues. In updateMessage, you have reverted to a previous version of firebase storage, probably an error arisen from erroneous merging with the main branch. The second request I have to ask is that in DataConverterTest, a number of variables are reused in multiple tests, such as `big_long`. I would love it if you could concentrate all of these at the top of the file.

oliveja22 (approved):
Thanks for changing the pull request as I requested, it seems to be quite good now, the variables are nicely stored at the top of the file and the firebase location for the data message is now correct.

PR #279, opened 2024-11-28
Feature implmenting profil picture in each screen that need it

TheoSchlaeppi:
This PR introduces the implementation of profile pictures in the `NewUser`, `UserProfile`, and `ListChat` screens.

- **NewUser Screen**: Users can add a profile picture while creating their account. It is also possible to skip adding a picture.
- **UserProfile Screen**: Users can view their profile picture and update it by clicking on it.
- **ListChat Screen**: Users can see the profile pictures of all their contacts. If a contact does not have a picture, a placeholder will be displayed.

### Tests

Tests have been added, but they do not include taking a photo directly as this is already covered in the tests for the `PhotoRequester` feature. Adding such tests here would result in unnecessary duplication.


matiisalv (approved):
Excellent job as always!! Very clear and well-writtenn, i have some recommendation , could you in the future add more documentation at your code.Other of that all is perfect for me .

PR #283, opened 2024-11-28
Feature/matiisalv/archiving books

matiisalv:
This PR implement the logic to archived Book in antoher collection for archiving , in BookFirestoreSource


app/src/main/java/com/android/bookswap/data/source/network/BookFirestoreSource.kt
```diff
@@ -252,6 +259,156 @@ class BooksFirestoreSource(private val db: FirebaseFirestore) : BooksRepository
...
+              "genres" to book.genres.map { it.toString() },
+              "userId" to book.userId.toString(),
+              "archived" to true,
+              "exchange" to book.exchange)
+
+      performFirestoreOperation(
+          db.collection(collectionBooks_Archived).document(book.uuid.toString()).set(bookMap)) {
+              result ->
+            if (result.isSuccess) deleteBooks(book.uuid, book, callback)
+            callback(result)
```
oliveja22, line 288: Here you call delete the book but you do not readd it to the other side, i.e you remove it from the non-archived archive and fail to retrieve it. Additionally there is no test for this function, it should be tested if only shoddily.

oliveja22 (changes requested):
The pull request is mostly good enough in quality and content, however there is one function that should be looked at again. That is moveBookToArchive, as for one it is not tested anywhere, and it should be tested. Additionally, there is the fact that you delete the book from the non-archived but you don't add it to the archived section.
Another issue is that you regenerate and recreate bookSource for every test, this should be created at the top of the file. The same can be said uuid you create every time.

app/src/main/java/com/android/bookswap/data/source/network/BookFirestoreSource.kt
matiisalv: This part actually add the book

oliveja22 (approved):
Thank you for taking into account all changes, as a tip for next time, always go one by one making sure functions are closed where they should be closed to avoid the issue we were having. Other than that, good work.

oliveja22 (changes requested):
This request changes is just to avoid people merging too early. I made the mistake of approving too early under the promise that tests would be added. But we fear someone else may merge ahead of time, hence the new request changes.

oliveja22 (approved):

PR #284, opened 2024-11-29
fix:git problem

matiisalv:
this pr comes to fix problem with git problem

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/map/Map.kt
+++ b/app/src/main/java/com/android/bookswap/ui/map/Map.kt
@@ -92,21 +92,21 @@ fun MapScreen(
   val cameraPositionState = rememberCameraPositionState()
   // Get the user's current location
   val latitude = geolocation.latitude.collectAsState()
-    val context = LocalContext.current
-    var isOnline = remember { isNetworkAvailable(context) }
+  val context = LocalContext.current
+  var isOnline = remember { isNetworkAvailable(context) }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -89,13 +92,20 @@ fun MapScreen(
...
+  val context = LocalContext.current
+  var isOnline = remember { isNetworkAvailable(context) }
   val longitude = geolocation.longitude.collectAsState()
   // Start location and books updates
   LaunchedEffect(Unit) {
-    bookManagerViewModel.startUpdatingBooks()
+    if (isOnline) {
+      bookManagerViewModel.startUpdatingBooks()
+    } else {
+      Toast.makeText(context, "Please connect to Internet to actualise", Toast.LENGTH_SHORT).show()
```
oliveja22, line 103: I think the word you are looking for here is update or even show, if possible please change the word "actualise", as in english that means to make something come true, to make it into reality, which is way off from what is here looked for.

oliveja22 (approved):
The content of this pull request seems to be quite good, in fact, it is extraordinary in the way that the connectivity is found, but most essential is the tests; it tests all possible situations which I can see, giving very reliable testing for the future. There are a few typos but we can chalk that up to bad luck, definitely not something worth not approving over.

PR #289, opened 2024-11-30
Fix the navigation to ensure that you can't go back to the auth route.

TheoSchlaeppi:
This PR fixes an issue where a new user navigating back using the `goBack` function on the Map screen could return to the `NEW USER` or `Login` screen if the previous route was `AUTH`. This behavior was unintended and has now been corrected.

In addition to this fix:
1. A new test has been added to ensure the `goBack` function prevents navigation back to `AUTH`.
2. The old `goBack` test has been removed as it is now obsolete due to the changes in functionality.

This change ensures a smoother and more intuitive navigation experience for new users.


LeniCadorinEpfl (approved):
Great work! Thank you very much for this very useful PR. Everything looks good to me. You can definitely merge it.

PR #292, opened 2024-12-02
Changes to How Messages Are Stored in Firebase

oliveja22:
# Changes to How Messages Are Stored in Firebase
This pull request seeks to update the methods and structure by which the messages are stored in the Firestore Database

## Previous Storage Method
Previously, all messages were stored in a single collection (messages) without subdivision. Each message was represented as a document, as can be seen below:
``` css
messages/
  <message_uuid>: {
    DataMessage
  }
``` 
## Updated Storage Method
Messages are now subdivided into conversation-specific subcollections based on the participants' UUIDs. The subcollections are identified by the concatenation of the sender's and receiver's UUIDs in size order. The new structure can be seen below:
``` css
chats/
  <sorted_uuid_1>_<sorted_uuid_2>/
    messages/
      <message_uuid>: {
        DataMessage
      }
``` 
## Changes to files
Of course, the file by which all of these interactions with firebase are performed has been updated in its entirety along with all relevant tests. Most notably `getMessages` is the function that has changed the most, as now it has to search for a particular set of messages and not all messages at once. The function `sendMessage` has also seen significant conceptual changes to ensure the new structure is always maintained. All other functions include minor fixes to adapt to the new structure.

Additionally, there has also been changes to `updateMessageBoxMap`, which has not only been changed to use the new `getMessages` function, but as an added bonus, also filters out all `IMAGE` type messages to ensure only `TEXT` messages are displayed in the message box.

`ChatScreen` has also seen changes to adapt to the new functions which could at time require additional parameters.

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
+++ b/app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
@@ -11,13 +11,14 @@ import com.google.firebase.firestore.FirebaseFirestore
 import com.google.firebase.firestore.ListenerRegistration
 import java.util.UUID
 
-const val COLLECTION_PATH = "messages"
+const val COLLECTION_PATH = "chats"
 /**
  * Firestore source for managing messages.
  *
  * @param db The Firestore database instance.
  */
 class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageRepository {
+
   /**
    * Generates a new UUID.
    *
@@ -26,6 +27,7 @@ class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageReposit
   override fun getNewUUID(): UUID {
     return UUID.randomUUID()
   }
+
   /**
    * Firestore source for managing messages.
    *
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/bookswap/data/source/network/MessageFirestoreSource.kt
```diff
@@ -157,95 +193,117 @@ class MessageFirestoreSource(private val db: FirebaseFirestore) : MessageReposit
...
+        .document(message.uuid.toString())
+        .get()
+        .addOnCompleteListener { task ->
+          if (task.isSuccessful && task.result != null && task.result!!.exists()) {
+            val document = task.result!!
+            val existingMessage = documentToMessage(document).getOrNull()
+
+            if (existingMessage != null) {
+              val currentTime = System.currentTimeMillis()
+              val fifteenMinutesInMillis = 15 * 60 * 1000
```
TheoSchlaeppi, line 262: This could be a const

TheoSchlaeppi (changes requested):
Thank you for this PR, which clearly optimizes information fetching and message storage. Overall, the code is excellent. However, I noticed a minor value that could be transformed into a constant, and I flagged it in the thread.  

Additionally, passing the context as a parameter to use Toasts raises some concerns for me. Generally speaking, it’s cleaner to display a Toast directly within the UI layer's callback rather than passing the context as a parameter to a ViewModel function (e.g., in `MessageFirestoreSource.kt`).  

That said, I don’t believe that a simple Toast would create the same safety issues that this pattern could introduce, such as memory leaks. These typically arise when context is used differently or retained incorrectly. However, if you also feel this code could be improved in this direction, I’ll leave that decision to you. Otherwise, the constant is the only change that’s strictly important.  

All the tests have been updated beautifully, and you’ve also updated `ContactVM`, which interacts heavily with the changes introduced in the repository.  

In summary, great work! Please address the small issues mentioned, and feel free to reach out as soon as you’ve made the corrections.

TheoSchlaeppi (approved):
Looks good for me !

PR #293, opened 2024-12-03
Fix ISBN check

oliveja22:
## Fix ISBN check
This pull request is quite small and just changes the place where the ISBN check is called to ensure it is not checked at every input, which previously resulted in not being able to input anything unless copying and pasting a valid ISBN in one go.


matiisalv (approved):
all good for me

PR #298, opened 2024-12-04
Fix Listener for Chat

oliveja22:
# Fix Listener for Chat
## Fix the Listener
In this pull request the Listener has officially and finally been corrected and is in full working fashion, given the change of structure to how the messages were stored in the firebase, it made the function to add the listener much easier to do. Now there is in `ChatScreen` a `DisposableEffect` in which the function is called with an appropriate callback, in this case storing the messages for offline usage.
## Convert to MVVM structure
Additionally, in this pull request all functionality which could have been moved to a different MVVM file has been moved. There is now a new file, `ChatScreenViewModel`, and its respective tests. It includes the functionality of adding both users to each other's contact list whenever the first message is sent as well as the timestamp formatting. This means that there is no more call to firebase directly from `ChatScreen`, it is now in `ChatScreenViewModel`. All of this is then called in the `LaunchedEffect`.


Chocolat2 (approved):
Overall changes are good and properly documented.
Some parts could be rewritted to comply with the best practices for android development and to improve code style and quality but this implementation works fine as is.

PR #300, opened 2024-12-05
Implement translation from geocoords latitude and longitude to an address

Chocolat2:
Implement translation from geocoords latitude and longitude to an address
Cleanup of PR #277

### DIFF
--- a/app/src/main/java/com/android/bookswap/MainActivity.kt
+++ b/app/src/main/java/com/android/bookswap/MainActivity.kt
@@ -57,6 +57,8 @@ import com.android.bookswap.ui.profile.NewUserScreen
 import com.android.bookswap.ui.profile.OthersUserProfileScreen
 import com.android.bookswap.ui.profile.UserProfile
 import com.android.bookswap.ui.theme.BookSwapAppTheme
+import com.android.bookswap.utils.ManifestUtils
+import com.google.android.libraries.places.api.Places
 import com.google.firebase.Firebase
 import com.google.firebase.auth.auth
 import com.google.firebase.firestore.FirebaseFirestore
@@ -90,6 +92,10 @@ class MainActivity : ComponentActivity() {
 
     // Initialize the geolocation
     val geolocation = Geolocation(this)
+    val apiKey = ManifestUtils.getApiKeyFromManifest(this)
+    if (!Places.isInitialized() && apiKey != null) {
+      Places.initialize(applicationContext, apiKey)
+    }
     BookSwapAppTheme {
       // A surface container using the 'background' color from the theme
       Surface(
@@ -99,6 +105,7 @@ class MainActivity : ComponentActivity() {
                 messageRepository = messageRepository,
                 bookRepository = bookRepository,
                 userRepository = userDataSource,
+                startDestination = C.Route.AUTH,
                 photoStorage = photoStorage,
                 messageStorage = messageStorage,
                 geolocation = geolocation,
@@ -226,7 +233,13 @@ class MainActivity : ComponentActivity() {
           }
         }
         navigation(startDestination = C.Screen.USER_PROFILE, route = C.Route.USER_PROFILE) {
-          composable(C.Screen.USER_PROFILE) { UserProfile(photoStorage) }
+          composable(C.Screen.USER_PROFILE) {
+            UserProfile(
+                photoStorage,
+                topAppBar = { topAppBar("Your Profile") },
+                bottomAppBar = { bottomAppBar(this@navigation.route ?: "") },
+            )
+          }
           composable(C.Screen.BOOK_PROFILE) { backStackEntry ->
             val bookId = backStackEntry.arguments?.getString("bookId")?.let { UUID.fromString(it) }
 
--- a/app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
+++ b/app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
@@ -97,26 +97,34 @@ class BookManagerViewModel(
     var successUsers = false
     var currentAttempt = 0
     while ((!successBooks || !successUsers) && currentAttempt < MAXIMUM_RETRIES) {
-      userRepository.getUsers { users ->
-        if (users.isSuccess) {
-          _allUsers.value = users.getOrNull()!!
-          successUsers = true
-        } else {
-          Log.e("BookManagerViewModel", "Failed to fetch users.")
+      val userJob = coroutineScope {
+        launch {
+          userRepository.getUsers { users ->
+            if (users.isSuccess) {
+              _allUsers.value = users.getOrNull()!!
+              successUsers = true
+            } else {
+              Log.e("BookManagerViewModel", "Failed to fetch users.")
+            }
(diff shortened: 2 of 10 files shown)


Chocolat2:
Tests still need to be updated to take into account the changes made in this PR

app/src/main/java/com/android/bookswap/utils/ManifestUtils.kt
```diff
@@ -0,0 +1,20 @@
+package com.android.bookswap.utils
+
+import android.content.Context
+import android.content.pm.PackageManager
+
+object ManifestUtils {
+  // Function to retrieve the API key from the manifest file
+  fun getApiKeyFromManifest(context: Context): String? {
```
EdenKahane, line 8: Why use this instead of `BuildConfig.API_KEY` like in ChatGPTApiService or the Map ? 
It's best if every API key is at the same place.

app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
```diff
@@ -97,26 +97,34 @@ class BookManagerViewModel(
     var successUsers = false
     var currentAttempt = 0
     while ((!successBooks || !successUsers) && currentAttempt < MAXIMUM_RETRIES) {
-      userRepository.getUsers { users ->
-        if (users.isSuccess) {
-          _allUsers.value = users.getOrNull()!!
-          successUsers = true
-        } else {
-          Log.e("BookManagerViewModel", "Failed to fetch users.")
+      val userJob = coroutineScope {
```
EdenKahane, line 100: Same as before, while this is a good change, usually it's best to keep it in a separate PR.

app/src/main/java/com/android/bookswap/model/map/Geolocation.kt
```diff
@@ -39,11 +42,9 @@ class Geolocation(private val activity: Activity) : IGeolocation {
 
   /** Location request settings */
   private val locationRequest: LocationRequest =
-      LocationRequest.create().apply {
-        interval = 10000 // Update interval in milliseconds
-        fastestInterval = 5000 // Fastest update interval in milliseconds
-        priority = LocationRequest.PRIORITY_HIGH_ACCURACY
-      }
+      LocationRequest.Builder(Priority.PRIORITY_HIGH_ACCURACY, LOCATION_REFRESH_DELAY)
```
EdenKahane, line 45: Great idea removing magic numbers !

EdenKahane (changes requested):
Really great code ! 
The overall PR is perfect, just try to really separate the changes in different PR.
There shouldn't be a BookManagerViewModel modification in a geocoords PR.
But this is minor as you already cleaned the PR quite a lot.

The only thing that really need to be fixed is the API Key retrieval.

Chocolat2:
> Really great code ! The overall PR is perfect, just try to really separate the changes in different PR. There shouldn't be a BookManagerViewModel modification in a geocoords PR. But this is minor as you already cleaned the PR quite a lot.

Right, ***IF*** the `BookManagerViewModel` was working properly in the first place, I wouldn't have to fix it so that I dont have endless network calls slowing down the entire app. Thus preventing me from moitoring the efficiency of my implementation of the LatLng to address translation feature as well as improving the geolocation efficiency to recieve a valid location faster so that the map isn't always showing (0'0N, 0'0E) upon launch.
I do agree however that, in general, I my PRs can be broken down in smaller PRs and still need to improve this.


> The only thing that really need to be fixed is the API Key retrieval.

What kind of problem did you encounter with this part ? It does work fine on my part.

Chocolat2:
> Really great code ! The overall PR is perfect, just try to really separate the changes in different PR. There shouldn't be a BookManagerViewModel modification in a geocoords PR. But this is minor as you already cleaned the PR quite a lot.

Right, ***IF*** the `BookManagerViewModel` was working properly in the first place, I wouldn't have to fix it so that I dont have endless network calls slowing down the entire app. Thus preventing me from moitoring the efficiency of my implementation of the LatLng to address translation feature as well as improving the geolocation efficiency to recieve a valid location faster so that the map isn't always showing (0'0N, 0'0E) upon launch.
I do agree however that, in general, I my PRs can be broken down in smaller PRs and still need to improve this.


> The only thing that really need to be fixed is the API Key retrieval.

What kind of problem did you encounter with this part ? It does work fine on my part.

app/src/main/java/com/android/bookswap/MainActivity.kt
```diff
@@ -226,7 +233,13 @@ class MainActivity : ComponentActivity() {
           }
         }
         navigation(startDestination = C.Screen.USER_PROFILE, route = C.Route.USER_PROFILE) {
-          composable(C.Screen.USER_PROFILE) { UserProfile(photoStorage) }
+          composable(C.Screen.USER_PROFILE) {
+            UserProfile(
+                photoStorage,
+                topAppBar = { topAppBar("Your Profile") },
```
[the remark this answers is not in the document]
Chocolat2: Right, next time, I'll *fix* some code to be able to work more efficiently, *revert* the fix before making the PR, before creating  new task that *may or may not* be assigned, so that the code I *already fixed* can be fixed...
Doing the work twice isn't efficient at all, and delaying bug-fixing is not a good practice either.

An alternative method to that would be to branch out of the current feature branch, make a PR to merge the fix in the feature branch and finally have a PR to merge the feature and the fix to main.

EdenKahane:
I really fail to see how the broken part of BookManagerViewModel affect a PR about geolocation. 
If something broken is affecting the PR you are implementing, sure fix it. 
But don't start fixing other bugs you noticed that shouldn't affect the geolocation.

When you notice an error, create a new task and see with the Scrum Master if you can solve this and when will it be solved.
Don't start fixing everything in one big PR, this lose the whole point of doing easily check-able and reversible part of the code, SCRUM is not used so everyone can just do one "everything-in-one" PR.

As for the API KEY see this message : https://github.com/BookswapEPFL/Bookswap/pull/300#discussion_r1872486485
On this one I'm not sure, but from what I'm seeing your using another system than what we're all using to retrieve the API KEY you need.

app/src/main/java/com/android/bookswap/MainActivity.kt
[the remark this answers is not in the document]
EdenKahane: Already answered here : https://github.com/BookswapEPFL/Bookswap/pull/300#issuecomment-2526202093

Is something is blocking you from working, sure fix it.
But don't go into a file trying to solve every single issue, this is not the point of your task nor your PR and remove the whole point.
Adding a topAppBar and bottomAppBar is not something that was discussed with the team and certainly not something we all agree on. I really doubt that missing the "Your profile" at the top was something making you less productive.

We cannot have each of your PR starting to fix something that we didn't assign before when it's not related at all ! The point of having a SM and tasks is to have a general feeling of what people are going to work on.
You deciding that your task on geocoords should also include the unnecessary and probably unwanted need to add a top and bottom bar is not normal.

EdenKahane (approved):
All the changes are mostly related to this PR.
So I'm approving it. Thank you for the change.

PR #303, opened 2024-12-06
Fix adding book from ChatGPT and directly moves it to EditScreen

EdenKahane:
Adding a book using a photo and sending it to ChatGPT was not working.
This was mainly due to the API KEY name change and the ISBN containing '-' characters.

Additionally, once the photo is retrieved, it should be sent to the EditScreen instead of just getting added. This necessitated to move the simple Book retrieval logic inside the screen instead of in the navigation (where it shouldn't be).

For sonarcloud : https://github.com/BookswapEPFL/Bookswap/pull/303#issuecomment-2522034830


EdenKahane:
Again, for SonarCloud, I'm not adding any line. I mainly modifying or removing already existing lines ! 
Just by virtue of big refactoring, some lines are noted as "new" even though they are not.

matiisalv (approved):
Excellent work as usual, i have nothing to say

PR #323, opened 2024-12-08
Add Button to Other User Profile that Goes to Chat

oliveja22:
## Add Button to Other User Profile that Goes to Chat
This pull request is quite simple but necessary, it adds a button to the other user profile such that when pressed it sends you directly to a chat with said user. Whether or not the user is in your contacts already makes little difference as that is handled by the `ChatScreen`. There is not much more to say other than a small test was modified to add a verification that the button exists and is clickable.


TheoSchlaeppi (approved):
Thank you for this small addition. Although it wasn't difficult to implement, it was essential for the proper functioning of the application. I approve this change without any further comments.

PR #324, opened 2024-12-09
Add Delete Functionality with URL and Photo ID Support

TheoSchlaeppi:
This PR enhances the `PhotoFirebaseStorageSource` by introducing improved deletion capabilities and tests.

#### Summary of Changes
- Added `deletePhotoFromStorage` function to delete a photo using its `photoId`.
- Added `deletePhotoFromStorageWithUrl` function to delete a photo using its URL.
- Extracted the `photoId` from URLs using regex to support deletion via URL.
- Included unit tests to verify both successful and failure scenarios for both deletion methods.

The addition of these methods provides flexibility to delete photos either by their unique ID or directly via their URL. This improves usability and makes the photo deletion API more robust and versatile.

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/repository/PhotoFirebaseStorageRepository.kt
+++ b/app/src/main/java/com/android/bookswap/data/repository/PhotoFirebaseStorageRepository.kt
@@ -8,4 +8,7 @@ interface PhotoFirebaseStorageRepository {
 
   fun addPhotoToStorage(photoId: String, bitmap: Bitmap, callback: (Result<String>) -> Unit)
   // getphoto is useless (as we can use the url to directly retrieve the picture and show it )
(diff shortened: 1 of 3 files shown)


LeniCadorinEpfl (changes requested):
Great job on this PR! The enhancements you’ve introduced to the PhotoFirebaseStorageSource are well thought out and provide significant value to the code by increasing flexibility and robustness. 

Here is just little suggestion for improvement:

Currently, if the regex fails to extract the photoId, it passes a null string to deletePhotoFromStorage, which could lead to unintended behavior or errors.

**Suggestion:** Add a check for matchResult and handle cases where the photoId is not extractable, maybe by calling the callback with a failure result (e.g., Result.failure(IllegalArgumentException("Invalid photo URL"))).

Overall this PR is really nice, well written and really useful, addressing the above suggestion would just make this PR even stronger.
Thank you for your hard work!

PR #325, opened 2024-12-09
Fix a bug where the selection of a filter with "-" was not selected 

TheoSchlaeppi:
This PR fixes a bug where filters containing a hyphen "-" were no longer selected when returning to the screen. The issue was due to the comparison logic not handling special characters correctly, leading to discrepancies when updating the selected genres.

To address this, the code was updated to normalize the genre strings by removing hyphens and ensuring case-insensitive comparisons. This change ensures that genres such as "Science-Fiction" and "SCIENCEFICTION" are properly recognized and retained.

With this fix, the filter functionality should now work as expected, preserving selected genres, even those with special characters.

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/DataBook.kt
+++ b/app/src/main/java/com/android/bookswap/data/DataBook.kt
@@ -40,8 +40,7 @@ enum class BookLanguages(val languageCode: String) {
   ENGLISH("English"), // English language
   SPANISH("Spanish"), // Spanish language
   ITALIAN("Italian"), // Italian language
-  ROMANSH("Romansh"),
-  SMALLTEST("Small Test"), // Romansh, a language spoken in Switzerland
+  ROMANSH("Romansh"), // Romansh, a language spoken in Switzerland
   OTHER("Other") // All languages that are not yet implemented
 }
 /** Genre of a book */
--- a/app/src/main/java/com/android/bookswap/data/DataBook.kt
+++ b/app/src/main/java/com/android/bookswap/data/DataBook.kt
@@ -40,7 +40,8 @@ enum class BookLanguages(val languageCode: String) {
   ENGLISH("English"), // English language
   SPANISH("Spanish"), // Spanish language
   ITALIAN("Italian"), // Italian language
-  ROMANSH("Romansh"), // Romansh, a language spoken in Switzerland
+  ROMANSH("Romansh"),
+  SMALLTEST("Small Test"), // Romansh, a language spoken in Switzerland
   OTHER("Other") // All languages that are not yet implemented
 }
 /** Genre of a book */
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/bookswap/model/map/BookFilter.kt
```diff
@@ -50,8 +50,9 @@ class BookFilter {
   fun setGenres(genres: List<String>) {
     val newGenresFilter =
         genres.mapNotNull { genre ->
-          // Use the Genre parameter to get the corresponding BookGenres enum
-          BookGenres.values().firstOrNull { it.Genre.equals(genre, ignoreCase = true) }
+          BookGenres.values().firstOrNull {
+            it.Genre.replace("-", "").equals(genre.replace("-", ""), ignoreCase = true)
```
Chocolat2, line 54: This is not where the problem happens, if the genre `SCHOOLMANUAL("School Manual")` is added, the bug will happen again.
The problem stems from the lines §123-124 of the `FilterMap.kt` file where filters are set again but using the enum class members names instead of their `Genre` value.

Chocolat2 (changes requested):
This change doesn't fix the actual problem and removes an explanation comment.

app/src/main/java/com/android/bookswap/model/map/BookFilter.kt
TheoSchlaeppi: Thank you for your help. My PR fixed the issue with the dash, but I didn't realize that the bug could also occur with spaces. My next commit should resolve this issue.

app/src/main/java/com/android/bookswap/ui/map/FilterMap.kt
Chocolat2, line 1: Good fix on the genres, however the problem still exists on the languages

app/src/main/java/com/android/bookswap/data/DataBook.kt
Chocolat2, line 1: Changes in this file can be reverted since these were only used for testing purposes.

app/src/main/java/com/android/bookswap/model/map/BookFilter.kt
```diff
@@ -64,7 +64,8 @@ class BookFilter {
   fun setLanguages(languages: List<String>) {
     val newLanguagesFilter =
         languages.mapNotNull { language ->
-          runCatching { BookLanguages.valueOf(language.uppercase()) }.getOrNull()
+          // Use the Genre parameter to get the corresponding BookGenres enum
```
Chocolat2, line 67: Small copy-paste typo on the comment, should be:
```
// Use the languageCode parameter to get the corresponding BookLanguages enum
```

app/src/main/java/com/android/bookswap/ui/map/FilterMap.kt
Chocolat2: New changes address this issue.

Chocolat2 (changes requested):
Two small problems to be fixed:
- test language added in the `DataBook.kt` file should be removed
- the comment on line 67 in the `BookFilter.kt` file has a copy-paste typo that should be fixed

PR #327, opened 2024-12-09
Book Profile Screen Fix

oliveja22:
# Book Profile Screen Fix
This pull request seeks to completely fix and finalise the Book Profile Screen as the title indicates, this includes a number of changes which will be described in parts below. Most changes are in `BookProfileScreen.kt`, however other files have been modified and added. Of course all changes come with their necessary tests. Two small fixes that don't deserve a part is adding the top app bar and removing some unnecessary spaces.
## Image Display
The first part to be fixed was the image display. Of old it simply held a placeholder and was prepared to consist of a list of images instead. Now it displays the image belonging to the book being observed so long as it exists, and reverts to a placeholder whenever it does not exist. This can be seen in the first of the images below.
## Synopsis Box Fix
This part is more complex, in order to not allow a particularly long synopsis to eat up the entire screen, it has now been made so it can hold a maximum of 15 lines (subject to change as it is a variable at the top of the file) and any more will allow the box to be scrolled through to reach the rest of the synopsis. This makes it much more visible and is observable in the second image below. To do this a new file was created, `DrawVerticalScrollBar.kt`, this file contains a new modifier that can be called to create a scroll bar to the right of the composable. It can be used as follows:
``` kotlin
val scrollState = rememberScrollState()
//Content
Modifier
  .verticalScroll(scrollState)
  .drawVerticalScrollbar(scrollState)
//Content
```
This is just an example but most parameters of the scroll bar can be modified by passing them as parameters, though they have a default value if a parameter is not passed.
## Edit Button Fix and Other User Button
This part of the pull request seeks to fix the edit button which was done changing the call to obtain the correct UUID as well as adding the Other User Button which allows someone to go to the user profile of the book owner if it is not their own. This was a small and fast change done by changing the `OnClick()` and `Text` of the edit button depending on the owner.
## Images
<div style="display: inline-block;">
  Screenshot1
  Screenshot2
  Screenshot3
</div>

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/components/DrawVerticalScrollBar.kt
+++ b/app/src/main/java/com/android/bookswap/ui/components/DrawVerticalScrollBar.kt
@@ -1,0 +1,43 @@
+package com.android.bookswap.ui.components
+
+import androidx.compose.foundation.ScrollState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.drawWithContent
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.StrokeCap
+import com.android.bookswap.ui.theme.ColorVariable
+
+const val STANDARD_THICKNESS = 4f
+const val MINIMUM_COERCION = 1f
+const val END_PADDING = 8f
+
+/** Modifier that draws a vertical scrollbar on the right side of the content. */
+fun Modifier.drawVerticalScrollbar(
+    scrollState: ScrollState,
+    color: Color = ColorVariable.Accent,
+    thickness: Float = STANDARD_THICKNESS,
+    paddingEnd: Float = END_PADDING
+): Modifier =
+    this.then(
+        Modifier.drawWithContent {
+          drawContent()
+
+          // Only draw the scroll bar if the content is scrollable
+          if (scrollState.maxValue > 0) {
+            // Height of the scrollbar (fixed proportion of the total height)
+            val scrollBarHeight = size.height * 0.1f
+            // Scrollbar position based on scroll progress
+            val scrollBarY =
+                (size.height - scrollBarHeight) *
+                    (scrollState.value /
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/bookswap/ui/components/DrawVerticalScrollBar.kt
```diff
@@ -0,0 +1,51 @@
...
+    color: Color = ColorVariable.Accent,
+    thickness: Float = STANDARD_THICKNESS,
+    paddingEnd: Float = END_PADDING
+): Modifier =
+    this.then(
+        Modifier.drawWithContent {
+          drawContent()
+
+          // Only draw the scroll bar if the content is scrollable
+          if (scrollState.maxValue > 0) {
```
matiisalv, line 35: could a hard code value

app/src/main/java/com/android/bookswap/ui/components/DrawVerticalScrollBar.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.foundation.ScrollState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.drawWithContent
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.StrokeCap
+import com.android.bookswap.ui.theme.ColorVariable
+
+/** Constants for the scrollbar */
+const val STANDARD_THICKNESS = 4f
```
matiisalv, line 12: maybe could add more specification of about the value means

matiisalv (approved):
Wonderful work! I think you did a very good job modeling your code and enhancing the tests for the Book Profile screen, especially fixing the issue with app configuration in the tests. Apart from the minor fixes I mentioned, which are episodic, you can merge your great work.

app/src/main/java/com/android/bookswap/ui/components/DrawVerticalScrollBar.kt
oliveja22: This is not something that could be modified later or something, it is simply checking that it is positive.

PR #329, opened 2024-12-10
Fix app crash on first launch after login

Chocolat2:
Fix app crash when navigating to the Map screen for the first time and location permissions are not yet granted by the user when location service enters the composition

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
@@ -1,6 +1,5 @@
 package com.android.bookswap.ui.map
 
-import android.Manifest
 import android.content.Context
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.geometry.Offset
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/android/bookswap/ui/map/MapScreenTest.kt
```diff
@@ -145,224 +140,247 @@ class MapScreenTest {
...
+            .assertCountEquals(2)
+        composeTestRule
+            .onAllNodesWithTag(C.Tag.BookDisplayComp.author, useUnmergedTree = true)
+            .assertCountEquals(2)
+        composeTestRule
+            .onAllNodesWithTag(C.Tag.BookDisplayComp.rating, useUnmergedTree = true)
+            .assertCountEquals(2)
+        composeTestRule
+            .onAllNodesWithTag(C.Tag.BookDisplayComp.filled_star, useUnmergedTree = true)
+            .assertCountEquals(9)
```
oliveja22, line 198: These numbers could perhaps use the variables set above, so the sum of the ratings for the filled stars, the size of the list of books for all above the rating, etc.. This would ensure that if someone changes the dataBook list due to adding more tests, it still works. But I don't think this is important enough.

oliveja22: I am also not certain that it is of your own doing, as it is marked as green but there are massive deletions and additions, so it may just be github playing and it was not you to begin with.

PR #330, opened 2024-12-10
Fix Get User

oliveja22:
This pull request is very simple, there was a small error in the getUser in the file `UserFirestoreSource.kt` which impeded properly getting users. It now works.


Chocolat2 (approved):
Good fix!

PR #332, opened 2024-12-11
Remove useless argument in deleteBooks in book repository

Chockencore:
This PR comes from the closed PR : https://github.com/BookswapEPFL/Bookswap/pull/251.
The other PR was closed because it was full of bugs and became far too big as the merges were added each week, so it was decided to split the tasks into several smaller PR.

This PR removes an argument from the BookRepository's deleteBooks function. The uuid argument was sufficient for this function, so the argument that wanted the whole book was removed.

### DIFF
--- a/app/src/main/java/com/android/bookswap/data/repository/BookRepository.kt
+++ b/app/src/main/java/com/android/bookswap/data/repository/BookRepository.kt
@@ -49,8 +49,7 @@ interface BooksRepository {
    * Function to delete a book from the repository.
    *
    * @param uuid The unique identifier of the book to be deleted.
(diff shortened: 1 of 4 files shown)


LeniCadorinEpfl (approved):
Thank you very much for your work! Your fix looks great. This PR demonstrates a clear and focused approach to enhancing the deleteBooks function by simplifying its signature and removing the unnecessary dataBook parameter. I will approve this PR, but as I wanted to say and Gregory just commented it before me, it could be better to use the name `DeleteBook` instead of `DeleteBooks` but as this is not a major issue I can already accept this PR such that you can quickly change the name and then merge.

PR #334, opened 2024-12-12
Create view model for edit and add book

Chockencore:
This PR comes from the closed PR : https://github.com/BookswapEPFL/Bookswap/pull/251.
The other PR was closed because it was full of bugs and became far too big as the merges were added each week, so it was decided to split the tasks into several smaller PR.

This PR removes a part of the code that was in AddToBookScreen and move it to a dedicated viewModel to adhere to the MVVM architecture. It was necessary to also create one for the EditToBookScreen as it used a part of the removed code from the AddToBookScreen.
This PR didn't change anything in the screen otherwise as they will be modified in this task https://github.com/BookswapEPFL/Bookswap/issues/249, even if the arguments given to the method of the viewModel are a bit clunky. The two screen will be standardized with a common component in the other task.

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/books/add/AddToBook.kt
+++ b/app/src/main/java/com/android/bookswap/ui/books/add/AddToBook.kt
@@ -1,8 +1,6 @@
 package com.android.bookswap.ui.books.add
 
-import android.content.Context
 import android.util.Log
-import android.widget.Toast
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
@@ -27,8 +25,6 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
 import com.android.bookswap.data.BookGenres
 import com.android.bookswap.data.BookLanguages
-import com.android.bookswap.data.DataBook
-import com.android.bookswap.data.repository.BooksRepository
 import com.android.bookswap.model.InputVerification
 import com.android.bookswap.model.LocalAppConfig
 import com.android.bookswap.model.add.AddToBookViewModel
@@ -42,7 +38,6 @@ import com.android.bookswap.ui.MAXLENGTHTITLE
 import com.android.bookswap.ui.components.ButtonComponent
 import com.android.bookswap.ui.components.FieldComponent
 import com.android.bookswap.ui.theme.ColorVariable.BackGround
-import java.util.UUID
 
 private const val HORIZONTAL_PADDING = 30
 /**
@@ -239,18 +234,18 @@ fun AddToBookScreen(
                           .fillMaxWidth(0.5f),
                   enabled = title.isNotBlank() && isbn.isNotBlank(),
                   onClick = {
-                      viewModel.saveDataBook(
-                          context,
-                          title,
-                          author,
-                          description,
-                          rating,
-                          photo,
-                          selectedLanguage?:BookLanguages.OTHER,
-                          isbn,
-                          selectedGenre?.let { listOf(it) }?: emptyList(),
-                          archived = false,
-                          exchange = true)
+                    viewModel.saveDataBook(
+                        context,
+                        title,
+                        author,
+                        description,
+                        rating,
+                        photo,
+                        selectedLanguage ?: BookLanguages.OTHER,
+                        isbn,
+                        selectedGenre?.let { listOf(it) } ?: emptyList(),
+                        archived = false,
+                        exchange = true)
                   }) {
                     Text("Save")
                   }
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/android/bookswap/ui/books/add/AddToBook.kt
```diff
@@ -238,53 +234,18 @@ fun AddToBookScreen(
                           .fillMaxWidth(0.5f),
                   enabled = title.isNotBlank() && isbn.isNotBlank(),
                   onClick = {
-                    // Check if title and ISBN are not blank (required fields)
-                    if (title.isNotBlank() &&
-                        isbn.isNotBlank() &&
-                        inputVerification.testIsbn(isbn) &&
-                        selectedGenre != null) {
```
oliveja22, line 245: Here you removed the verifications, you shouldn't, it is true that two of them are redundant with the enabled field, but the other two should be added.

app/src/main/java/com/android/bookswap/ui/books/edit/EditBookScreen.kt
```diff
@@ -270,59 +269,20 @@ fun EditBookScreen(
...
+                              photo,
+                              enumValues<BookLanguages>().firstOrNull { it.name == language }
+                                  ?: BookLanguages.OTHER,
+                              book.isbn ?: "",
+                              selectedGenre?.let { listOf(it) } ?: emptyList(),
+                              book.archived,
+                              book.exchange)
                         },
                         modifier = Modifier.fillMaxWidth().testTag(C.Tag.EditBook.save),
                         enabled = title.isNotBlank(),
```
oliveja22, line 288: Here you also need to add some conditions in enabled.

app/src/test/java/com/android/bookswap/model/edit/EditBookViewModelTest.kt
```diff
@@ -0,0 +1,181 @@
...
+          },
+          any())
+    }
+
+    verify { navigation.goBack() }
+  }
+
+  @Test
+  fun `updateDataBook don't navigate back on failure`() {
+    every { booksRepository.updateBook(any(), any()) } answers
```
oliveja22, line 111: You created this every twice, perhaps do it in the setup.

app/src/test/java/com/android/bookswap/model/edit/EditBookViewModelTest.kt
```diff
@@ -0,0 +1,181 @@
...
+                updatedBook.exchange == book.exchange
+          },
+          any())
+    }
+
+    verify { navigation.goBack() }
+  }
+
+  @Test
+  fun `updateDataBook don't navigate back on failure`() {
```
oliveja22, line 110: it should say doesn't

app/src/test/java/com/android/bookswap/model/edit/EditBookViewModelTest.kt
```diff
@@ -0,0 +1,181 @@
...
+        book.description!!,
+        book.rating!!.toString(),
+        book.photo!!,
+        book.language,
+        book.isbn!!,
+        book.genres,
+        book.archived,
+        book.exchange)
+
+    verify {
```
oliveja22, line 130: this verification is a bit unnecessary, as it was already tested before.

app/src/test/java/com/android/bookswap/model/edit/EditBookViewModelTest.kt
```diff
@@ -0,0 +1,181 @@
...
+          },
+          any())
+    }
+
+    verify(exactly = 0) { navigation.goBack() }
+  }
+
+  @Test
+  fun `deleteBook deletes book successfully`() {
+    val uuid = UUID.randomUUID()
```
oliveja22, line 151: maybe create this uuid in the setup rather than twice.

app/src/test/java/com/android/bookswap/model/edit/EditBookViewModelTest.kt
```diff
@@ -0,0 +1,181 @@
...
+        {
+          val callback = secondArg<(Result<Unit>) -> Unit>()
+          callback(Result.failure(Exception("Deletion failed")))
+        }
+
+    // Act
+    viewModel.deleteBook(context, uuid)
+
+    // Assert
+    verify { booksRepository.deleteBook(uuid, any()) }
```
oliveja22, line 178: this verification is nice but unnecessary, as it was already tested before

oliveja22 (changes requested):
The pull request is quite nice and overall important, however there are a few things to note. In the tests you sometimes create variables twice rather than in the setup, it is the case for a UUID in EditBookViewModelTest. Additionally, you sometimes unnecessarily verify a number of things twice, it is nice, but unimportant. Lastly, and perhaps most importantly, in AddToBook and EditBookScreen you removed a number of verifications that should be added again. I think what we discussed, and adding it in enabled is a good idea, but it has to be added somewhere. An example would be the ISBN test. Otherwise it is quite good work, and I am glad you divided it into smaller tasks, it also makes our lives as reviewers easier. It is also nice how it was modified to include a better MVVM architecture and most importantly, its newfound simplicity.

Chockencore:
I did the requested changes and modified a little bit the mainActivity to put the viewModels directly inside the parameter of the screen to avoid having too much line in the mainActivity. I also did it for the viewModels of the other screens.

oliveja22 (approved):
Everything looks good, thank you for taking into account my requests, so long as you get the CI working it should be fine.

PR #335, opened 2024-12-12
Update upload-artifact of github actions to v4

EdenKahane:
For several weeks, github announced the deprecation of upload-artifact@v3 announcing temporary disruption to warn people and later a full shutdown of the oldest version.
To allow long-term usage of this code (and prevent mail spam) an update was needed to v4.

This turned out to be very straightforward.


LeniCadorinEpfl (approved):
Thank you very much for this PR, this is very useful. You can definitely merge it.

PR #336, opened 2024-12-12
Fix crash when trying to resolve address with NaN coords

Chocolat2:
Fix crash happening while trying to translate an address from coordinates when location is not enabled and no previous location is available, thus using NaN values and causing a crash.


Chockencore (approved):
The fix works and is clean. The removal of the unnecessary computation of the camera position at launch is a nice addition.

PR #337, opened 2024-12-12
Book created from IBSN are sent to EditScreen

EdenKahane:
Before, when creating a book using the ISBN book creation screen, the books were directly added to the user list. 
Usually users were expecting to each time go search their book to edit it and add the genre and the language after.
Now when the book is added, the user is directly sent to the EditScreen where the book can be edited.

To note, I also fixed the problem in the ISBN creation where the userId of the createdBook was not always the user creating the book.

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/ui/books/add/ISBNAddTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/ui/books/add/ISBNAddTest.kt
@@ -22,7 +22,6 @@ import com.android.bookswap.model.LocalAppConfig
 import com.android.bookswap.model.UserViewModel
 import com.android.bookswap.resources.C
 import com.android.bookswap.ui.navigation.NavigationActions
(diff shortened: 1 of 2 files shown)


Chockencore (changes requested):
This Pr is good and it is nice to remove the toast as it would be a bit invasive when arriving on the edit screen.
But I have a question, how does this code reflect this part of the PR :
"To note, I also fixed the problem in the ISBN creation where the userId of the createdBook was not always the user creating the book."
The only change about uuid is replacing "appConfig.userViewModel.uuid" by "appConfig.userViewModel.getUser().userUUID", which should be always equal.
As I didn't have the bug inside this branch or the main branch, I assume that it was already patched.
It would be better to go back on this change as it was more efficient before.

EdenKahane:
@oliveja22 told me that he didn't think it was solved, and he solved it using that.

PR #340, opened 2024-12-12
Modify user's bookList when adding in addToBookScren and deleting book in EditBookScreen

Chockencore:
This PR add the functionality of adding automatically the book to the current user's book list when adding a new book in the app and removes it from his book list when deleting it. It uses the userViewModel to update the user's booklist.


TheoSchlaeppi (approved):
Thank you for this addition. The code looks well-written, and the tests have been updated accordingly. I don’t have any further suggestions. For me, it’s approved.

PR #341, opened 2024-12-12
Fix adding book from ISBN not fetching the correct language

EdenKahane:
Someone changed languageCode so it's not a code anymore but just a language name, this broke data retrieval from GoogleBooks.
I renamed the now changed languageCode to languageName (because it's just that) and created a new (but actually just the old deleted one) languageCode, and now everything works.


matiisalv (approved):
all good for me

PR #342, opened 2024-12-13
Feature/matiisalv/disconnect

matiisalv:
This PR the option to be disconnected to the application and repear to the original Screen without being connected


app/src/main/java/com/android/bookswap/model/userViewModel.kt
```diff
@@ -157,4 +159,14 @@ open class UserViewModel(
...
   }
+  /**
+   * Disconnect the current user by resetting the user data and UUID. This method sets the user data
+   * to a new DataUser instance with a new UUID, marks the user as not loaded, and updates the
+   * stored state to false.
+   */
+  fun disconnectUser() {
+    this.dataUser = DataUser(UUID.randomUUID())
+    this.uuid = UUID.randomUUID()
+    Firebase.auth.signOut()
```
Chockencore, line 170: The comment says that it marks the user as not loaded but the function doesn't change the isLoaded or the _isStored. Also it should update the user if the _isStored is false before disconnecting the user, to avoid lose of data.

Chockencore (changes requested):
The PR is fine but it lacks some tests, the disconnect function doesn't do what it says it is doing and the it should update the user if _isStored is false.
Also, the disconnect button wasn't modified to show what the disconnect button should look like.
Right now, the button isn't the same size of the edit button, and I think that it would be better to put the disconnect button in place of the profile icon at the top right with a icon of shutting down. But the figma should be modified and approved before adding the new disconnect button.

app/src/main/java/com/android/bookswap/ui/profile/UserProfile.kt
```diff
@@ -262,7 +262,16 @@ fun UserProfile(
...
               }
-
+          // Disconnect Button
+          Button(
+              onClick = {
+                appConfig.userViewModel.disconnectUser()
+                Toast.makeText(context, "Disconnected", Toast.LENGTH_SHORT).show()
+                navigationActions.navigateTo(C.Screen.AUTH)
+              },
+              modifier = Modifier.align(Alignment.CenterHorizontally)) {
```
Chockencore, line 272: A testTag should be added to the button to do some tests

app/src/main/java/com/android/bookswap/model/userViewModel.kt
Chockencore: Also the dataUser should take in argument the new "this.uuid" that is created with the randomUUID(), right now they use different UUID which is strange.

PR #352, opened 2024-12-17
Strings.xml and C.kt

TheoSchlaeppi:
This PR moves all user-facing strings to `strings.xml` or `C.kt`. This includes:
- Text displayed in the UI
- Toast messages
- Enum values such as language and genre

Strings used in logs have not been moved, as keeping them in the code simplifies debugging and they are not visible to the user. Similarly, strings used for content descriptions remain unchanged, as they are also not directly visible to the user.

If I missed any user-facing strings, please feel free to flag them.

### DIFF
--- a/app/src/main/java/com/android/bookswap/MainActivity.kt
+++ b/app/src/main/java/com/android/bookswap/MainActivity.kt
@@ -12,6 +12,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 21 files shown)


oliveja22 (changes requested):
It all looks very good, however there are two particular strings which would be great to be added to the strings file despite not being directly linked to the UI or anything the user sees. These are the very large strings we have on two occasions, one in particular I believe should be applied here. That is the chatGPT prompt. I believe a string as large as it should be in the strings file as it makes modifying it easier. There is also a fairly large one in BookProfileScreenTest, but given that that one is on me and that it is a test, it is not as important as the chatgpt one. Otherwise I believe it is great work, and if there are more strings, I failed to find them as well.

oliveja22 (approved):
Thanks for taking into account my request.

PR #354, opened 2024-12-17
Fix Automatic Login Issue

oliveja22:
## Fix Automatic Login Issue
There was an error that oft occurred when login in past the first one where the uuid would not correctly load. It has now been fixed and it works without fall.


TheoSchlaeppi (approved):
Thank you for fixing this issue!

PR #356, opened 2024-12-17
Add tests coverage for PhotoRequester

EdenKahane:
PhotoRequester is a hard class to test because it relies on permissions and ActivityResultContracts.
The best way to test that I found is following this : https://medium.com/@adrian.gl/how-to-mock-and-test-activityresult-apis-with-jetpack-compose-f3a1d6311171
But I added more type check safety because otherwise finer testing was not possible.

### DIFF
--- a/app/src/androidTest/java/com/android/bookswap/model/PhotoRequesterTest.kt
+++ b/app/src/androidTest/java/com/android/bookswap/model/PhotoRequesterTest.kt
@@ -13,7 +13,6 @@ import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.graphics.ImageBitmap
-import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.core.app.ActivityOptionsCompat
 import androidx.core.content.FileProvider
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/com/android/bookswap/model/PhotoRequesterTest.kt
```diff
@@ -0,0 +1,245 @@
...
+import java.io.File
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertThrows
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+/*
+Thanks https://medium.com/@adrian.gl/how-to-mock-and-test-activityresult-apis-with-jetpack-compose-f3a1d6311171
```
matiisalv, line 34: could add explanation about it  in DocString

matiisalv (changes requested):
Amazing work as usual , but could a comment in the rest  of the code , because give explanation in a certain part ,other than that it 's perfect, good work .

app/src/androidTest/java/com/android/bookswap/model/PhotoRequesterTest.kt
EdenKahane: The function that I documented is a complicated helper function. But this is just a test class like all the other so the documentation is not needed.

matiisalv (approved):
as we discuss all good

PR #357, opened 2024-12-18
Feature let people putting books without isbn

TheoSchlaeppi:
This PR let user adding books without ISBN in the rare case where a book that they want to exchange has no isbn. It modify only the bookrepository. Feel free to suggest me another way to do it if you think there is optimization possible.


LeniCadorinEpfl (approved):
Thank you very much for your work. It really improves the book addition process and looks ready for merging from my perspective.

PR #358, opened 2024-12-18
Fix map marker display

Chocolat2:


oliveja22 (approved):
I think this pull request is quite good, especially if it manages to fix that longstanding issue we have been facing and postponing... The simplification of BookFirestoreSource seems so incredible, it is almost excessive; I am not one to complain about it though. There are no hardcoded values, I see sufficient documentation and nothing stands out as worrisome. Tests for this should already exist so the absence of modifications for any test file feels justified, especially given sufficient coverage according to SonarCloud. I therefore have no reason to not approve it.

PR #360, opened 2024-12-18
Improve add and edit screens

Chockencore:
This PR uses the newly implemented component that displays the list of fields, in the AddToBookScreen and EditBookScreen.  By doing that a lot of old tests were obsolete, a few were updated and some were removed as they were redundant with the tests of the EntriesListBookTest.
This PR also modifies the component EntriesListBookComponent :
1) the photo fields to a button that redirects to the camera. It uses the already implemented ButtonIcons and added some new functionnality to it like the possiblity to choose the right icon and put in its own file.
2) The fields have now a maximum of lines to avoid a field that covers all the screen.  Also to avoid having an circle in place of a rectangle with smooth corner when there are multiples lines, the shapes of the FieldComponent were a little bit modified. When there is a single line, the shape is the same as the old one.
3) The unique fields like the photo button and the stars were moved to the bottom of the list for a better look.
4) The stars at the end of the text for the fields were removed for the ones that aren't necessary.


TheoSchlaeppi (changes requested):
This PR is very good and the implementation is smooth. There's a minor detail that needs to be changed: the typo in 'pics' should be corrected to 'Take a photo' (I flagged this in the thread). Additionally, can you please center the stars component horizontally?

Aside from these minor adjustments, I believe the code is ready for approval.

TheoSchlaeppi (approved):
Looks. good for me !

PR #361, opened 2024-12-18
Add Address Modification

oliveja22:
## Add Address Modification
This pull request is much larger than was initially expected, mostly due to the size of the app at this given instance. It seeks to introduce in every necessary place a series of text fields that would allow any user to modify their address, that is the address that all other people would see when they go into the `OtherUserProfile` page. The fields that it adds can be seen below, in the visible image. Additionally, it can be noted that it is a component that has been made separately and used in two different places. The component is `AddressFieldsComponent` and it is used in `EditProfile` and `NewUser`. It has two options, in `NewUser` proportions are slightly different and country and city are mandatory. In edit profile unless country and city are filled in it does nothing new but allows the user to continue as they could have modified other fields. The address is transformed into a latitude and longitude and that is how it is stored. There are two options, a comprehensive and detailed list of values that becomes an accurate longitude and latitude or a smaller list of just city and country. This allows people who want more privacy that option. It stores latitude and longitude nonetheless, at the center of that city. 
Another addition is that in the `OtherUserProfile`, one can now see the city and country of the person that owns the profile that is being observed.
All necessary tests have been modified as necessary and new ones added (a harder task than one might have expected).
Image of new fields specifically in `EditProfile` in this case:
image


oliveja22:
As a comment that I should add, the large size of the PR is due to two main factors. First ktfmt check shifted a number of things in the edit user screen, creating a disproportionate amount of changes, resulting in 300 lines rather than the more accurate 40-50. On top of that, tests account for almost 900 lines, this is due to a different way of testing and a number of empty lines in between to ensure readability. This means the amount of lines modified is disproportionate to the actual content of the pr.

matiisalv (approved):
Very goof job as usual, well written, and structured and documented  , thanks for that , but  a major  problem is that that PR way to long , in future project please if it possible to make various PR for such big work, other than that your work is wonderful and help a lot the time

matiisalv:
> As a comment that I should add, the large size of the PR is due to two main factors. First ktfmt check shifted a number of things in the edit user screen, creating a disproportionate amount of changes, resulting in 300 lines rather than the more accurate 40-50. On top of that, tests account for almost 900 lines, this is due to a different way of testing and a number of empty lines in between to ensure readability. This means the amount of lines modified is disproportionate to the actual content of the pr.

I just see your comment no comment so on , perfect work

PR #362, opened 2024-12-19
Fix map display of User's self-location marker

LeniCadorinEpfl:
### Summary
This PR addresses the issue in the `MapScreen` where the user had both a blue dot and a marker for his location. The changes ensure the marker:
- Only the blue dot is always present
- When click near your location (50-meter radius) the Marker shows and the Marker's text directly appears with "Your Location"
- When you click on the map outside of the 50 meter distance from the location, the Marker is no more displayed.

### Key Changes
**1. Map Screen :**
- Marker Visibility State:
  - Added a showMarker state variable to control the marker's visibility dynamically.
- Marker State Management:
  - Introduced MarkerState to manage the position of the user's self-location marker.
  - Used LaunchedEffect to update the marker position when the latitude/longitude changes.
- Distance Calculation for Clicks:
  - Calculated the distance between the clicked point and the user's location using Location.distanceBetween.
  - Displayed the marker only if the clicked point is within a 50-meter radius.
- InfoWindow Behavior:
  - Showed or hid the InfoWindow based on the visibility of the marker using markerState.showInfoWindow() and markerState.hideInfoWindow().

**2. UI Tests:** 
- Added a test markerNotDisplayedWhenLaunch to verify:
  - The marker is not displayed initially when the screen launches.

### UI visual representation:
Screenshot 2024-12-19 095721
Screenshot 2024-12-19 095740

### DIFF
--- a/app/src/main/java/com/android/bookswap/ui/map/Map.kt
+++ b/app/src/main/java/com/android/bookswap/ui/map/Map.kt
@@ -238,10 +238,7 @@ fun MapScreen(
               }
 
               GoogleMap(
-                  onMapClick = { // mutableStateSelectedUser = NO_USER_SELECTED
-                      // showMarker = true // Hide the marker when map is clicked elsewhere
-                      // Log.d("MapScreen", "Map clicked, hiding marker")
-                      clickedLatLng ->
+                  onMapClick = { clickedLatLng ->
                     Log.e("MapScreen", "Map clicked")
 
                     // Check if the clicked location is near the user's location
@@ -283,8 +280,7 @@ fun MapScreen(
                       title = stringResource(R.string.map_screen_your_location),
                       visible = showMarker,
                       icon =
-                          BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_BLUE),
-                  )
+                          BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_BLUE))
                 }
                 filteredUsers.value
                     .filter {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -210,8 +217,47 @@ fun MapScreen(
...
+                  markerState.showInfoWindow()
+                } else {
+                  markerState.hideInfoWindow()
+                }
+              }
+
               GoogleMap(
-                  onMapClick = { mutableStateSelectedUser = NO_USER_SELECTED },
+                  onMapClick = { clickedLatLng ->
+                    Log.e("MapScreen", "Map clicked")
```
TheoSchlaeppi, line 242: shouldn't it be Log.i ?

app/src/main/java/com/android/bookswap/ui/map/Map.kt
```diff
@@ -227,9 +273,11 @@ fun MapScreen(
               ) {
                 // Marker for user's current location
                 if (!latitude.value.isNaN() && !longitude.value.isNaN()) {
+                  Log.e("MapScreen", "Marker displayed at: ${latitude.value}, ${longitude.value}")
```
TheoSchlaeppi, line 276: I think Log.i would be better here too

TheoSchlaeppi (changes requested):
This PR is great, I tested it directly on the emulator and it works perfectly. There are two minor typos to fix: use 'log.i' instead of 'Log.e'. Otherwise, the code is good. Please let me know once these changes are made.

TheoSchlaeppi (approved):
Looks good for me !

PR #363, opened 2024-12-19
Feature add a toast when editing or adding a new book when no wifi

TheoSchlaeppi:
This PR introduce Toast to the editing and adding books screen that notify the user that the action will be perfomed when the wifi is connected. I use for this the function : isNetworkAvailable created in a older PR.


Chockencore (approved):
This PR looks fine to me, it's a nice feature to let the user know how we handle books when offline.

PR #365, opened 2024-12-19
Update README.md

LeniCadorinEpfl:
Update the `README` accordingly with the app newest features and MS3 deliverables.


README.md
```diff
@@ -1,30 +1,27 @@
...
-
-[Team minutes](https://docs.google.com/spreadsheets/d/1YF34CWge25Y1hSQsP5UWHCDoyYNJaxIPasSb13gWYao/edit?usp=sharing)
-
-[Architecture Diagram](https://lucid.app/lucidchart/a07f4bb7-a054-40e9-95be-d4cbb0485822/edit?viewport_loc=-1324%2C33%2C6059%2C2938%2C0_0&invitationId=inv_d502a57d-8d89-4e37-9a4f-f06ec2cf1941)
+## Introduction
+Welcome to BookSwap. The application consists of an online version of the free book exchange boxes often found in small villages scattered throughout the whole of the Swiss countryside, it helps avoid having to search for said box exchanges and additionally allows to search for a more desirable book than what perhaps would be in the box if one were to simply walk up to said box and have a look.
+
+## Features
+- **Adding Books:**  BookSwap offers three user-friendly methods to add books to your account. The first is the traditional approach, where you manually enter the book details. Alternatively, you can simply input the ISBN for quick addition. Lastly, BookSwap leverages an advanced AI model to extract book information automatically from a photo of the book, making the process seamless and efficient.
```
oliveja22, line 7: Here we have to make sure that it says the picture has to be taken of the back cover, the front cover will simply not work

oliveja22 (changes requested):
There is a small change that has to be made, mostly because if anyone reads the readme and attempts to take a picture to add their book, they will be met with utter disappointment, as one's first instinct is to take a picture of the front page, not the back cover. We need to point out that one must take a picture of the back cover specifically.

oliveja22 (approved):

PR #366, opened 2024-12-19
InfoWindow is displayed instead of navigating to user profile

Chockencore:
This PR removes the problem where clicking on a marker would navigate to the user profile linked to the marker. It was supposed to do the navigation only when doing a long press but was instead put in the onClick which was already used to display the infoWindow. As the Marker, doesn't have a modifier, the button to go to the user profile was put inside the InfoWindow as a new box.
This PR also removed the unused DisplayStarReview composable.


EdenKahane (approved):
Great PR that remove outdated stuff and fix annoying problem.
Keep up the great work

PR #367, opened 2024-12-19
Fix Login Issue

oliveja22:
This pull request fixes a small issue when login in that if it is too fast, it stops working.


LeniCadorinEpfl (approved):
Thank you for this very useful fix. You can definitely merge it !

PR #369, opened 2024-12-20
Fix users marker and book display

LeniCadorinEpfl:
This PR fixes an issue where the current user logged in marker was red as if he was another user and the user book where displayed as if it was other user's book.
This fix ensure the UI properly differentiates between the logged-in user and other users. And removes the logged-in user books in the list of books in the map.

### DIFF
--- a/app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
+++ b/app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
@@ -8,9 +8,9 @@ import com.android.bookswap.data.DataUser
 import com.android.bookswap.data.UserBooksWithLocation
 import com.android.bookswap.data.repository.BooksRepository
 import com.android.bookswap.data.repository.UsersRepository
+import java.util.UUID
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/bookswap/model/map/BookManagerViewModel.kt
```diff
@@ -113,7 +114,7 @@ class BookManagerViewModel(
         launch {
           booksRepository.getBook { books ->
             if (books.isSuccess) {
-              _allBooks.value = books.getOrThrow()
+              _allBooks.value = books.getOrThrow().filter { it.userId != currentUserUUID }
```
Chockencore, line 117: This is redundant with the line 168. The books with no users are filtered out, as the current user is removed from the allUser list, the books are removed at line 168

Chockencore (changes requested):
The PR is good, the .filter { it.userId != currentUserUUID } can be removed.

Chockencore (approved):
The PR is good.

PR #370, opened 2024-12-20
Fix navigation error to Chat

oliveja22:
This pull request seeks to fix the error that occurred when attempting to reach the chat from another user profile.


matiisalv (approved):
Very good job, thanks to optimize  the way to take the conctact

PR #375, opened 2024-12-20
Fix new breaking change to sign-android-release

EdenKahane:
No idea why, but since two days ago this bug started to affect the build apk ci... : https://github.com/ilharp/sign-android-release/issues/24

I can confirm that this solve the apk build error.


oliveja22 (approved):
Since it is quite simple, and if you can check that it works under every circumstance, I can see no reason to not approve it. The CI will do the rest of the work.

